PR #2, opened 2022-11-21
ObjectDumper: Replaced usage of the old 'GetEnumNames'

MostlyBored:
This fixes the Lua error `UEnum::GetEnumNames/GetNames is not safe for <4.15 because the names array type changed and we haven't abstracted that yet`.

This means the object dumper now works in UE 4.13.


narknon:
Thank you!

PR #6, opened 2022-11-23
Added a recursive tree-view for StructProperty values in LiveView.

MostlyBored:
The left offset relative to the owner of the struct.
The right offset is relative the struct itself.
The `Copy value` context menu option works identically to before, except you need to close the tree for the struct before the context menu appears when you right-click.

ue4ss_feature_showcase


MostlyBored:
Wrong branch.

PR #10, opened 2022-11-25
Hopefully removed the need for two UE4SS dlls.

MostlyBored:
Split 'FStaticMaterial' into two structs, one for <=4.19 and one for >=4.20.
This hopefully means that it's no longer necessary to have separate versions of UE4SS for <=4.19 and >=4.20.

### Note: This needs to be tested!

### DIFF
--- a/src/GUI/Dumpers.cpp
+++ b/src/GUI/Dumpers.cpp
@@ -13,6 +13,7 @@
 #include <UE4SSProgram.hpp>
 #include <Unreal/UnrealInitializer.hpp>
 #include <Unreal/Searcher/ObjectSearcher.hpp>
+#include <Unreal/NameTypes.hpp>
 #include <Unreal/UClass.hpp>
 #include <Unreal/AActor.hpp>
 #include <Unreal/UFunction.hpp>
@@ -23,10 +24,6 @@ namespace RC::GUI::Dumpers
 {
     using namespace ::RC::Unreal;
 
-    // TODO: Move these structs (and the enum) to a proper place in the Unreal module.
-    //       They need to be controlled by the engine version select or before 2.0 releases.
-
-    // In KH3, there's an extra float either at the end of FMeshUVChannelInfo or at the end of FStaticMaterial, I can't tell which one.
     enum { MAX_TEXCOORDS = 4, MAX_STATIC_TEXCOORDS = 8 };
     struct FMeshUVChannelInfo
     {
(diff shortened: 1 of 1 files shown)


narknon:
Thanks, was wondering if this was a possible solution but didn't have a chance to try it yet. I'll test tomorrow and probably do a new release soon, especially if this works.

MostlyBored:
This method is not ideal because we're having to duplicate code just to accommodate each version of the struct.
There's no avoiding multiple structs, but we can get rid of the code duplication.

The syntax could look something like this in the end:
```c++
// Put these defines somewhere handy, perhaps right after the structs.
// The order of the structs and expressions in these two defines must be identical.
#define FStaticMaterial_Versions FStaticMaterial_419AndBelow, FStaticMaterial_420AndAbove
// It's important that only one of these selectors evaluate to true at any given point.
#define FStaticMaterial_VersionSelectors Version::IsAtMost(4, 19), Version::IsAtLeast(4, 20)

// Usage whenever you want to access 'FStaticMaterial'.
ConditionalMultiStructExecute<FStaticMaterial_Versions>([&]<typename FStaticMaterial>(FStaticMaterial*) {
    // In this lambda, we can now use 'FStaticMaterial' as an agnostic type.
    // This code would expand to a bunch of if-statements.
    // Under the hood, 'FStaticMaterial' would refer to 'FStaticMaterial_419AndBelow' if
    // 'Version::IsAtMost(4, 19)' evaluates to true and 'FStaticMaterial_420AndAbove' if
    // 'Version::IsAtLeast(4, 20)' evaluates to true.
}, FStaticMaterial_VersionSelectors);
```

You could also make the syntax better with an additional macro:
```c++
// After the 'Versions' and 'VersionsSelectors' macros.
#define FStaticMaterial_ConditionalMultiStructExecute(Code) ConditionalMultiStructExecute<FStaticMaterial_Versions>([&]<typename FStaticMaterial>(FStaticMaterial*) { Code }, FStaticMaterial_VersionSelectors);

// Usage whenever you want to access 'FStaticMaterial'.
FStaticMaterial_ConditionalMultiStructExecute({
    // Same as the lambda in the previous example.
})
```

This would of course also be reusable for other structs with minimal effort, just create the structs and the defines.

The implementation for `ConditionalMultiStructExecute` would be something like this:
```c++
template<typename V>
void ConditionalMultiStructExecute_Internal(auto callable, [[maybe_unused]]int obj, bool expression)
{
    if (expression)
    {
        callable(std::bit_cast<V*>(nullptr));
    }
}

template<typename V, typename... Vs>
void ConditionalMultiStructExecute_Internal(auto callable, int obj, bool expression, auto... expressions)
{
    if (expression)
    {
        callable(std::bit_cast<V*>(nullptr));
    }
    ConditionalMultiStructExecute_Internal<Vs...>(callable, obj, expressions...);
}

template<typename V, typename... Vs>
void ConditionalMultiStructExecute(auto callable, bool expression, auto... expressions)
{
    ConditionalMultiStructExecute_Internal<V, Vs...>(callable, 0, expression, expressions...);
}
```

narknon:
Thank you, that is good to know for the future, until a potential fully dynamic method later on.  

You definitely don't need to and don't feel any pressure to do so, but if you'd be willing to look at the usmap implementation branch, it would be appreciated.  Shade gave permission to port it over, but I'm having weird issues with variables not linking correctly and things not being able to be casted that are casted the same way elsewhere.  I'm debating if it would be easier to duplicate the CXX gen and then prune it down to the required information for mappings files, rather than trying to port code made from a different UE templating system.

I know one function I added to FProperty isn't going to work the way I did it because the version specific generated files don't contain it.

PR #31, opened 2023-01-07
Added 'edit value' functionality to LiveView using the 'ImportText' virtual function.

UE4SS:


narknon:
Merged in through other commits.

PR #35, opened 2023-01-16
Add config = Engine flag for UCLASS default/project user configs

Buckminsterfullerene02:
As far as I can see this doesn't cause any conflicts, although I couldn't find any info about GlobalUserConfig needing Config = Engine in it, so I just left that one for now.


Buckminsterfullerene02:
I added this as a setting as requested. I attempted to build the xinput myself to test, but when loading the game there is an error `[17:32:55] Fatal Error: Call to K32GetModuleBaseNameA failed. Error Code: 6` but I don't have time right now to figure that one out.

Buckminsterfullerene02:
This seems to work as intended.

Issue #36, opened 2023-01-16
[2.2.0 UHT Generator] PublicDependencyModuleNames list is not populated correctly


Buckminsterfullerene02:
In the <name>.Build.cs file for the project, the `PublicDependencyModuleNames` array is not generated properly anymore.

For example, old version:
```cpp
PublicDependencyModuleNames.AddRange(new string[] {
            "AIModule",
            "AnimGraphRuntime",
            "AnimationCore",
            "AnimationSharing",
            "AudioMixer",
            "Chaos",
            "Core",
            "CoreUObject",
            "DLSSBlueprint",
            "DSTelemetry",
            "DeveloperSettings",
            "Engine",
            "FSDEngine",
            "GameplayTags",
            "GameplayTasks",
            "InputCore",
            "LevelSequence",
            "MediaAssets",
            "MediaUtils",
            "MovieScene",
            "MovieSceneTracks",
            "NavigationSystem",
            "Niagara",
            "NiagaraCore",
            "NiagaraShader",
            "OnlineSubsystem",
            "OnlineSubsystemUtils",
            "PhysicsCore",
            "PropertyPath",
            "Slate",
            "SlateCore",
            "UMG",
        });
```

New version:
```cpp
PublicDependencyModuleNames.AddRange(new string[] {
            "AIModule",
            "Core",
            "CoreUObject",
            "Engine",
            "FSDEngine",
            "GameplayTasks",
            "NavigationSystem",
            "Niagara",
            "PhysicsCore",
            "UMG",
        });
```

1. Use following settings:
```
IgnoreAllCoreEngineModules = 0
IgnoreEngineAndCoreUObject = 1
MakeAllFunctionsBlueprintCallable = 1
MakeAllPropertyBlueprintsReadWrite = 1
MakeEnumClassesBlueprintType = 1
```
2. UHT Dump
3. Fix all normal errors
4. Watch as your world burns with many headers missing errors

This issue is almost certianly cost by [this line of code](https://github.com/UE4SS-RE/RE-UE4SS/commit/eed11361a09c17c27366590f16f9cd6cff5917f7#diff-f2d1c353a4bdf4ce653bc93b980513944149dc55cbd2d6e5f220d6ea1c6a9692R1384) in #33  

According to the changelogs, this is only meant to be checked when the engine dumping is enabled, but the current code does not appear to actually check the state of that setting. 

[...]

PR #37, opened 2023-01-16
(#36) Fix UHT build.cs generation

praydog:
before

after
asdgadsg


Buckminsterfullerene02:
@praydog Well this is awkward. I tested the hotfix release (as well as my own changes that definately have this fix in it) and the public dependency module names list is still broken in the same way as it was. I have no idea how it worked for you!

PR #38, opened 2023-01-17
Fully conserve equality between generations by ordering #includes and class/struct pre-declerations

Buckminsterfullerene02:
Tested on DRG, works. However, I think that the commandlet also needs to be updated since it is responsible for generating the engine module includes. But everything on the UE4SS side is done.

Obviously I am making sure that CoreMinimal.h and the generated.h includes are kept at top and bottom of include list respectively.


Buckminsterfullerene02:
Tested properly with full build of new template instead of just up to the point of running the commandlet, all bugs fixed now, should be good to go.

PR #78, opened 2023-02-23
Fixed: Github action properly packages up and releases a nightly along with attaching the changelog to the release

razor950:
I swapped the changelog into markdown format so the workflow script would read it without any issues.

This needs the commit to be tagged with nightly for it to build and only applies to the 2.xDev branch at the moment, I can clean that up to only need a tag in the future.


UE4SS:
I think this is a good idea but was there any particular reason why you removed the other action ?
While nightly builds are cool, it's also useful to know that no commit breaks the build or to be notified when they do break the build, especially considering this nightly action must be manually triggered with a tag.

razor950:
oh, we can keep the other one to make sure builds don't break or I can swap this to not use tags and instead just build a specific branch every night and make a release.

UE4SS:
I think it might be wasteful to build every night since we don't have new commits every night.
I think we should keep the other action.

The tag thing should probably be kept unless either: A. activity increases to the point where there's always new commits every day, or, B. the action can check for new commits and only make a nightly in that case.
But of course, it wouldn't be terrible to be able to manually trigger a nightly build either so the tag thing might be nice either way.

PR #101, opened 2023-04-19
Cpp modding

localcc:
# C++ modding

This adds the ability to write C++ mods.

When a c++ mod is loaded, the loader searches for 2 functions - `start_mod` and `uninstall_mod`.

The first one should create an instance of `ICPPMod` and the second one should destroy it.

This is how an example implementation could look like:

```c++
#include <Mod/ICPPMod.hpp>

#define CPP_MOD_TEST_API __declspec(dllexport)

class TestMod : public RC::ICPPMod
{
public:
    TestMod() : ICPPMod()
    {}

    ~TestMod()
    {}

    auto update() -> void override
    {
    }
};

extern "C"
{
    CPP_MOD_TEST_API RC::ICPPMod* start_mod()
    {
        return new TestMod();
    }

    CPP_MOD_TEST_API void uninstall_mod(RC::ICPPMod* mod)
    {
        delete mod;
    }
}
```

The update function gets called every UE4SS update tick. 

For the mod to work and be loaded by UE4SS correctly, it needs to be linked in to the `cppsdk` library and built in the ue4ss codebase.

An example `CMakeLists.txt` could look like this:

```cmake
cmake_minimum_required(VERSION 3.18)

set(TARGET CppModTest)
project(${TARGET})

add_library(${TARGET} SHARED "dllmain.cpp")
target_link_libraries(${TARGET} PUBLIC cppsdk)
```

# Cpp sdk

Because the mod links to `cppsdk`, a `cppsdk.dll` file must be placed inside of the `Mods` folder, this allows for any injection method to be used for ue4ss (`ue4ss.dll` and `xinput1_3.dll`). However, two versions of `cppsdk` get compiled, `cppsdk.dll` and `cppsdk_xinput.dll`.

For the xinput version, the `cppsdk_xinput.dll` must be renamed to `cppsdk.dll` and put in the `Mods` folder, for the default version, the `cppsdk.dll` should be put in the `Mods` folder.

## Why is cppsdk needed?

Without the cppsdk the dll would not be able to get loaded on different injection methods, as the linker would search for the exact dll name that the mod was linked to, e.g. if the mod was linked to `ue4ss`, then a `ue4ss.dll` would have to be present in the process. 
Cppsdk dll serves a proxy role reexporting all `ue4ss.dll`/`xinput1_3.dll` symbols so that mods can get loaded regardless of the `ue4ss` injection method.

# Warning
When compiling mods, be sure to build them in the same configuration as the ue4ss file used, e.g. if ue4ss is compiled in Debug, the mod should be too, otherwise, the C runtime versions will not match, and it will break.


src/Mod/LuaMod.cpp
UE4SS, line 1: Could you undo all of the code-style changes ?
I think the easiest way this can be done is to save your non-style changes somewhere, remove the file, copy the old `Mod.cpp` file (the file itself, not the contents into a new document) and then rename the file to `LuaMod.cpp` and then apply your code changes.

localcc: Sure, is there a way to make the code style consistent with the rest of the codebase? I thought the .clang-format file would take care of that.

UE4SS: > Sure, is there a way to make the code style consistent with the rest of the codebase? I thought the .clang-format file would take care of that.

Unfortunately there isn't at the moment, not beyond just looking at existing code.
Clang-format was unable to satisfy me and I must've left the file in for possible future reference if they ever added the features that I wanted.
The clang-format file should probably have been removed a long time ago.

UE4SS:
I think perhaps `CppModule` should be renamed to `CppMod` to stay consistent with `LuaMod`.
Sorry if this sounds nitpicky but I think it really helps clarity if we keep to consistent naming conventions.

src/Mod/CppModule.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace RC
+{
+    CppModule::CppModule(UE4SSProgram& program, std::wstring&& mod_name, std::wstring&& mod_path) : Mod(program, std::move(mod_name), std::move(mod_path))
+    {
+        m_dlls_path = m_mod_path + L"\\dlls";
+
+        if (!std::filesystem::exists(m_dlls_path))
+        {
+            set_installable(false);
```
UE4SS, line 15: We should use `Output::send<LogLevel::Warning>(...)` here to let the user know why the dll couldn't be loaded.
In the future, this message could be sent via `set_installable` instead and outputted alongside the regular `Was unable to install mod` message.

src/Mod/CppModule.cpp
```diff
@@ -0,0 +1,60 @@
...
+            return;
+        }
+
+        m_main_dll_module = LoadLibraryW((m_dlls_path + L"\\main.dll").c_str());
+
+        if (!m_main_dll_module)
+        {
+            set_installable(false);
```
UE4SS, line 23: We should use `Output::send<LogLevel::Warning>(...)` here to let the user know why the dll couldn't be loaded.
In the future, this message could be sent via `set_installable` instead and outputted alongside the regular `Was unable to install mod` message.

src/Mod/CppModule.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+        m_start_mod_func = reinterpret_cast<start_type>(GetProcAddress(m_main_dll_module, "start_mod"));
+        m_uninstall_mod_func = reinterpret_cast<uninstall_type>(GetProcAddress(m_main_dll_module, "uninstall_mod"));
+
+        if (!m_start_mod_func || !m_uninstall_mod_func)
+        {
+            FreeLibrary(m_main_dll_module);
+            m_main_dll_module = NULL;
+
+            set_installable(false);
```
UE4SS, line 35: We should use `Output::send<LogLevel::Warning>(...)` here to let the user know why the dll couldn't be loaded.
In the future, this message could be sent via `set_installable` instead and outputted alongside the regular `Was unable to install mod` message.

src/Mod/CppModule.cpp
UE4SS: We should also use `GetLastError()` to further give the user more information.
It's probably fine to supply the error code and the user can then look up what that error code means.

localcc:
> I think perhaps `CppModule` should be renamed to `CppMod` to stay consistent with `LuaMod`. Sorry if this sounds nitpicky but I think it really helps clarity if we keep to consistent naming conventions.

I have thought of that, but that sounds like it implies it's deriving from `ICPPMod`

UE4SS:
Further information for usage:
A dll mod uses the same rules as a Lua mod.
That means that the dll must be named `main.dll` and reside in `Mods/<ModName>/dlls`.
There must be either an entry, `<ModName> : 1`, in `Mods.txt` or the file `enabled.txt` must exist in the `<ModName>` directory.

NOTE:
A cpp mod cannot exist alongside a Lua mod of the same name.
This is due to `find_mod_by_name` returning the first mod it finds.
It will either find the cpp mod or the Lua mod and whichever is first in the vector (load order) will get loaded.

UE4SS:
I have a few more thoughts:

1. I think `cppsdk.dll` should be named `UE4SS-cppsdk.dll`.
The reason for this is because I like the idea of being able to tell which files belong to UE4SS, partially for easy uninstallation but also to distinguish our files from other utilities.
There are other mod systems that use the same `Mods` directory so the fact that we put our dll in there isn't enough to fully distinguish in my opinion.

2. The `Unreal` module must have the changes from https://github.com/UE4SS-RE/RE-UE4SS/pull/101/commits/56daed2db24c82bde199d1f2c5f64d608158a77b applied and few other fixes.
I will do this after this pull request has been accepted.

Also I don't know if you saw my other reply about the usage of `Output::send` because it somehow got attached to the wrong commit so I'll link it here just in case: https://github.com/UE4SS-RE/RE-UE4SS/pull/101#discussion_r1171730646

src/Mod/CppModule.cpp
localcc: Oh I did not see that, I thought it was using the printf format because I saw it being used somewhere in `UE4SSProgram.cpp`

src/UE4SSProgram.cpp
```diff
@@ -750,11 +756,18 @@ namespace RC
     {
         Output::send(STR("Setting up mods...\n"));
 
-        if (!std::filesystem::exists(m_mods_directory))
+        if (!std::filesystem::exists(m_mods_directory)) { set_error("Mods directory doesn't exist, please create it: <%S>", m_mods_directory.c_str()); }
+
+        auto cpp_sdk_path = m_mods_directory / "cppsdk.dll";
```
UE4SS, line 761: Related to [570c637](https://github.com/UE4SS-RE/RE-UE4SS/pull/101/commits/570c637e2bd46884537ddba0221ca25a8b865342).
This should be renamed to `UE4SS-cppsdk.dll`.

src/UE4SSProgram.cpp
```diff
@@ -916,27 +930,11 @@ namespace RC
         for (auto& mod : m_mods)
         {
             // Remove any actions, or we'll get an internal error as the lua ref won't be valid
-            mod->clear_delayed_actions();
```
UE4SS, line 919: Was this intended ?
I'm not seeing any calls to `clear_delayed_actions` anywhere else and we need to clear the delayed actions when uninstalling.
I'm guessing this was meant to be moved into the `uninstall` function in the `LuaMod` class.

PR #103, opened 2023-04-23
Cmake fixes

trumank:
Prevents polyhook from `install`ing targets to fix complaints about polyhook being unable to find Zydis:

    CMake Error: install(EXPORT "PolyHook_2-targets" ...) includes target "PolyHook_2" which requires target "Zydis" that is not in any export set.
    CMake Error: install(EXPORT "PolyHook_2-targets" ...) includes target "PolyHook_2" which requires target "Zycore" that is not in any export set.

This definitely feels like a hack but I'm not familiar enough with the arcane magic of `cmake` to know for sure.

Also adds a few build related paths to .gitignore as they are creating noisy commits.


UE4SS:
Yes this definitely seems like a hack but it'll be good enough for now.
Thank you for the contribution.

PR #104, opened 2023-04-23
Remove stdin read which blocks until user input on Linux

trumank:
It's not clear to me what the purpose of this line is, but on Linux it will cause the game to appear to hang indefinitely without any user input in the console.


UE4SS:
This line of code has no purpose in this context.
Thank you for your contribution.

PR #106, opened 2023-04-24
Add Lua type definition generator and definitions for UE4SS built-ins

trumank:
https://user-images.githubusercontent.com/1144160/233905258-a7438c1b-8d67-4b99-8cf7-b19f883a2cd7.mp4

Adds Lua type definitions to be used by a [lua language server](https://marketplace.visualstudio.com/items?itemName=sumneko.lua).

[Annotation reference](https://github.com/LuaLS/lua-language-server/wiki/Annotations)

I decided to extend the existing CXXGenerator to generate Lua definitions as opposed to create an entirely new generator because they are more similar than they are different. I figured any quirks with how UE objects are accessed would be easier to keep updated if they weren't split between two different generators. The conditionals everywhere are undeniably ugly, however. Perhaps templates could be used to specify language at compile time? Or create intermediate structures to pass to specialized generators so the generators aren't operating directly on UE types? My experience with C++ is still quite limited so very welcome to feedback.

There are still many rough edges:
 - poor handling of properties and functions containing invalid characters
 - complex UE types
 - limited support of generics by the Lua LSP
 - probably many errors in my hasty, manual porting of the API document to Lua
 - need to implement helper functions with explicit types to get common objects
 
Usage is as simple as launching UE4SS and clicking the "Generate Lua Types" button which drops them into the `Mods/shared/types` directory. They will be picked up automatically by the LSP. Enums can be used too, but they must be `require()`'d so their values are available at runtime.

### DIFF
--- a/StagingDev/Mods/shared/Types.lua
+++ b/StagingDev/Mods/shared/Types.lua
@@ -1,0 +1,816 @@
+---@meta
+
+---@enum Key
+Key = {
+    LEFT_MOUSE_BUTTON = 0x1,
+    RIGHT_MOUSE_BUTTON = 0x2,
+    CANCEL = 0x3,
+    MIDDLE_MOUSE_BUTTON = 0x4,
+    XBUTTON_ONE = 0x5,
+    XBUTTON_TWO = 0x6,
+    BACKSPACE = 0x8,
+    TAB = 0x9,
+    CLEAR = 0x0C,
+    RETURN = 0x0D,
+    PAUSE = 0x13,
+    CAPS_LOCK = 0x14,
+    IME_KANA = 0x15,
+    IME_HANGUEL = 0x15,
+    IME_HANGUL = 0x15,
+    IME_ON = 0x16,
+    IME_JUNJA = 0x17,
+    IME_FINAL = 0x18,
+    IME_HANJA = 0x19,
+    IME_KANJI = 0x19,
+    IME_OFF = 0x1A,
+    ESCAPE = 0x1B,
+    IME_CONVERT = 0x1C,
+    IME_NONCONVERT = 0x1D,
+    IME_ACCEPT = 0x1E,
+    IME_MODECHANGE = 0x1F,
+    SPACE = 0x20,
+    PAGE_UP = 0x21,
+    PAGE_DOWN = 0x22,
+    END = 0x23,
+    HOME = 0x24,
+    LEFT_ARROW = 0x25,
+    UP_ARROW = 0x26,
+    RIGHT_ARROW = 0x27,
+    DOWN_ARROW = 0x28,
+    SELECT = 0x29,
+    PRINT = 0x2A,
+    EXECUTE = 0x2B,
+    PRINT_SCREEN = 0x2C,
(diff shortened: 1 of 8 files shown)


UE4SS:
I'm not a fan of squeezing in another generator like this.
Perhaps it would be a good idea to redesign CXXGenerator to be typed.
The new generator could have all the UE code and then each language could have its own specification for how to generate the type.

Here's a small example; I chose enums because it was nice and simple and I haven't looked into the other types to see if this is reasonable at all so feel free to give feedback
```c++
//Dumpers.cpp:
generate_cxx_headers(...);
generate_lua_types(...);

//Generator.hpp:
auto generate_cxx_headers() -> void;
auto generate_lua_types() -> void;

//Generator.cpp:
template<typename T>
class TypeGenerator
{
private:
    T specification{};

public:
    auto generate_enum() -> void
    {
        // Code here that's identical for all generators.
        // Things like casting UObject to UEnum and getting the names array from the enum.
	// Then finally we call into the specification to finish this type:
        specification.generate_enum(enum_names);
    }
};

class CXXHeaderGenerator
{
public:
    auto generate_enum() -> void
    {
        // C++ specific generation here.
    }
};

class LuaTypesGenerator
{
public:
    auto generate_enum() -> void
    {
        // Lua specific generation here.
    }
};

auto generate_cxx_headers() -> void
{
    TypeGenerator<CXXHeaderGenerator> generator{};
    generator.generate();
}

auto generate_lua_types() -> void
{
    TypeGenerator<LuaTypesGenerator> generator{};
    generator.generate();
}
```

trumank:
I replaced the language enum and branching with the template specification suggestion. It's a bit messy as I didn't want to risk breaking anything by trying to pull apart some of the function generation logic and let it spill into the specification for a first pass.

UE4SS:
This is looking good so far.
I'm ready to merge this after it's been tested unless you want to work on the second pass in the same pull request.
I think we should test this regardless if there's a second pass.
Perhaps generate before & after and confirm that the outputs are identical.
I believe the inconsistency between runs of the generator was fixed a while back, maybe @narknon can confirm that ?

trumank:
I have been comparing the full dump of DRG which is UE 4.27 as I go and so far it's identical, but I do not how much coverage that actually is.

narknon:
Yeah, Buck added alphabetical ordering to everything, so it should be consistent.

UE4SS:
I see you've made some changes.
I no longer know the status of this pull request, whether or not you want to continue working on it so I will leave it open until you comment letting me know it's all ready.

Also, if you want to refactor the CXX code (the second pass talked about earlier), feel free to do it in this pull request or another one later, either way it's fine as long as the generator output is identical to what was generated before you made any changes.
It should be trivial to check that and using DRG for testing is probably fine.
If anything still ends up being broken then we can get that fixed at that point.

trumank:
I'm satisfied with it for the time being. There are improvements to be made but I'll need to get a better feel for how the types work and they're already in a very usable state.

narknon:
Can you write something up in .md for the docs so people know what this is/how to use it?

PR #109, opened 2023-05-05
Add specifying of load order for mods used by BPModLoaderMod

Okaetsu:
Adds a way to specify in what order LogicMods should be loaded with BPModLoaderMod.

Load order of the mods can be specified in logicmods.txt found in the Mods folder. Example below:
```
SomeMod
AnotherMod
```
'SomeMod' will be loaded first and 'AnotherMod' will be loaded second, any mods that weren't included in the file will be loaded in random order _after_ SomeMod and AnotherMod.

This is useful in cases where a mod relies on another mod being loaded first, for example, utility mods.

### DIFF
--- a/Staging/Mods/BPModLoaderMod/Scripts/main.lua
+++ b/Staging/Mods/BPModLoaderMod/Scripts/main.lua
@@ -11,6 +11,10 @@ package.path = '.\\Mods\\ModLoaderMod\\?.lua;' .. package.path
 package.path = '.\\Mods\\ModLoaderMod\\BPMods\\?.lua;' .. package.path
 
 Mods = {}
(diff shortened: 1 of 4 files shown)


UE4SS:
Looks fine to me except the file name and location.
Can we rename `logicmods.txt` to `load_order.txt` and can we also put it inside the `Mods\BPModLoader` directory instead of the `Mods` directory ?
The reasoning here is that this file belongs to the BP mod loader and it only affects BP mods so putting it inside the `Mods` directory makes it very unclear as to what it does, especially considering the fact that "logic mod" is not a term exclusively used for BP mods and because the regular `mods.txt` file is also related to load-order.

Issue #110, opened 2023-05-05
CppMod ImGui interface


trumank:
CppMods should have a way to add a tab to the GUI for mod specific settings / widgets.

UE4SS:
I think this is a good idea.

Do you have any ideas for how we could deal with exporting ImGui to each C++ mod ?

The problem is that the ImGui context isn't preserved across dll boundaries so if we call an ImGui function from a C++ mod we'll call it on a nullptr ImGui context.

At the moment, the only solution I have is to call `ImGui::SetCurrentContext` & `ImGui::SetAllocatorFunctions` from each C++ mod.
Like this for example:
```c++
ImGui::SetCurrentContext(UE4SSProgram::get_current_imgui_context());
ImGuiMemAllocFunc alloc_func{};
ImGuiMemFreeFunc free_func{};
void* user_data{};
UE4SSProgram::get_current_imgui_allocator_functions(&alloc_func, &free_func, &user_data);
ImGui::SetAllocatorFunctions(alloc_func, free_func, user_data);
```
It would have to be part of some initialization (like `on_program_start`), and executed before any ImGui functions are called.
It's rather bulky and annoying to have to include this in every C++ mod but it might be acceptable (maybe with some refactoring to reduce & simplify) but obviously an alternative to this would be better.

trumank:
Ah, no I was not aware of the ImGui context issue. I imagine the bulk of it could be shoved into an `#include` so it's not necessary to copy and paste the same block.

So far I've been only thinking about how the actual interface would work. The most straightforward (stateless as far as UE4SS is concerned) would be to have two functions, one which returns whether the mod has a GUI (or an optional string with the GUI name) and one which is called to actually render the tab:

```C++
auto CppMod::gui_title() -> std::optional<std::wstring>; // none if no GUI; or wstring if GUI
auto CppMod::render_gui() -> void; // called by UE4SS when it's time to render the GUI
```

The other option would be to explicitly register GUI tabs/windows so it's possible to have many or none at all. The exact interface for this may depend on the ImGui docking branch which I've wanted to explore: https://github.com/ocornut/imgui/issues/2109

UE4SS:
> The other option would be to explicitly register GUI tabs/windows so it's possible to have many or none at all. The exact interface for this may depend on the ImGui docking branch which I've wanted to explore: [ocornut/imgui#2109](https://github.com/ocornut/imgui/issues/2109)

[...]

PR #111, opened 2023-05-07
Expose ImGui tab creation to CppMods

trumank:
I took a quick stab at #110. Some things to note:

1. I put `DebuggingGUI::m_tabs` behind a mutex to be safe, though this means tabs won't be able to add other tabs in the middle of a frame. Perhaps queuing new tabs up to add at the end of a frame would be better?
2. I moved mod setup to after console creation so the GUI would be available in the mod constructor. Assuming this doesn't break something else, this seems better than utilizing `on_program_start` as it is only run at UE4SS startup and doesn't run again for mod restarts.
3. I've included a hack that allows the mod .dll file to be swapped and reloaded, allowing hot reloading on my system using WINE. I thought very little about how sound it is and wouldn't actually include it, but it's nice for testing purposes.

I've got a test mod here: https://github.com/trumank/ue4ss-scratch/blob/20035c1dddd8b7f76686e408da34a8c2068dcab8/MyAwesomeMod/dllmain.cpp

### DIFF
--- a/src/GUI/GUI.cpp
+++ b/src/GUI/GUI.cpp
@@ -145,12 +145,14 @@ namespace RC::GUI
                     ImGui::EndTabItem();
                 }
 
-                std::lock_guard<std::mutex> guard(m_tabs_mutex);
-                for (const auto& tab : m_tabs)
                 {
-                    if (ImGui::BeginTabItem(to_string(tab->TabName).c_str()))
+                    std::lock_guard<std::mutex> guard(m_tabs_mutex);
+                    for (const auto& tab : m_tabs)
                     {
-                        tab->render();
+                        if (ImGui::BeginTabItem(to_string(tab->TabName).c_str()))
+                        {
+                            tab->render();
+                        }
                     }
                 }
 
--- a/src/GUI/GUI.cpp
+++ b/src/GUI/GUI.cpp
@@ -144,6 +144,16 @@ namespace RC::GUI
                     BPMods::render();
                     ImGui::EndTabItem();
                 }
+
+                std::lock_guard<std::mutex> guard(m_tabs_mutex);
+                for (const auto& tab : m_tabs)
+                {
+                    if (ImGui::BeginTabItem(to_string(tab->TabName).c_str()))
+                    {
+                        tab->render();
+                    }
+                }
+
                 ImGui::EndTabBar();
                 if (!is_unreal_initialized) { ImGui::EndDisabled(); }
             }
@@ -302,6 +312,17 @@ namespace RC::GUI
         m_os_backend->set_gfx_backend(m_gfx_backend.get());
         m_gfx_backend->on_os_backend_set();
         m_os_backend->on_gfx_backend_set();
+    }
+
+    auto DebuggingGUI::add_tab(std::shared_ptr<GUITab> tab) -> void
+    {
+        std::lock_guard<std::mutex> guard(m_tabs_mutex);
+        m_tabs.push_back(tab);
+    }
+    auto DebuggingGUI::remove_tab(std::shared_ptr<GUITab> tab) -> void
+    {
+        std::lock_guard<std::mutex> guard(m_tabs_mutex);
+        m_tabs.erase(std::remove(m_tabs.begin(), m_tabs.end(), tab), m_tabs.end());
     }
 
     DebuggingGUI::~DebuggingGUI()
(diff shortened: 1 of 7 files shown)


trumank:
https://user-images.githubusercontent.com/1144160/236652167-f49f7c2f-cf21-4043-803c-cf72ed9507b0.mp4

Oh, and some hot reload action. :D

src/GUI/GUI.cpp
```diff
@@ -144,6 +144,16 @@ namespace RC::GUI
...
                 }
+
+                std::lock_guard<std::mutex> guard(m_tabs_mutex);
+                for (const auto& tab : m_tabs)
+                {
+                    if (ImGui::BeginTabItem(to_string(tab->TabName).c_str()))
+                    {
+                        tab->render();
+                    }
+                }
```
UE4SS, line 155: You should wrap the mutex & for-loop in its own scope since we're only looking to protect `m_tabs` and not the entire `update` function:
```suggestion
                {
                    std::lock_guard<std::mutex> guard(m_tabs_mutex);
                    for (const auto& tab : m_tabs)
                    {
                        if (ImGui::BeginTabItem(to_string(tab->TabName).c_str()))
                        {
                            tab->render();
                        }
                    }
                }
```

UE4SS:
1. Can you separate the hot-reload hack into its own pull request ?
2. The reason why C++ mods are loaded as early as they are is because I want to give people as much opportunity as possible to do whatever they want to do and having C++ mods load as early as possible means they could potentially do things before UE inits like establish a proper lifetime cycle. Any delays make this more difficult as you decrease the time available for a mod to init before UE inits. Not sure how much time the output configuration takes but it's probably a low amount so your change is likely fine.

Some notes on the example mod:
1. The code that sets the context and allocator functions should probably be a macro supplied by UE4SSProgram.hpp.
2. You're using make_unique and then assigning the result to a shared_ptr. I'm guessing this was a simple mistake but I thought it was worth mentioning anyway.

trumank:
> Can you separate the hot-reload hack into its own pull request ?

Removed, I'll submit another PR shortly.

> The reason why C++ mods are loaded as early as they are is because I want to give people as much opportunity as possible to do whatever they want to do and having C++ mods load as early as possible means they could potentially do things before UE inits like establish a proper lifetime cycle. Any delays make this more difficult as you decrease the time available for a mod to init before UE inits. Not sure how much time the output configuration takes but it's probably a low amount so your change is likely fine.

Ah this makes sense. I'm not aware of how long the GUI takes to init, but if it's a problem, new tabs could be queued by `UE4SSProgram` so they don't have to wait. This brings up something I forgot to mention last night: Does the GUI ever not exist, or does it just hide when disabled? So far I haven't considered what happens if it's disabled in the config.

> You're using make_unique and then assigning the result to a shared_ptr. I'm guessing this was a simple mistake but I thought it was worth mentioning anyway.

It was my understanding that uniqe_ptr will cheaply and automatically get converted to another pointer type when necessary, so haven't been bothering to use anything else when creating a pointer. Is this not the case? Either way I guess it doesn't hurt to be explicit.

UE4SS:
> > The reason why C++ mods are loaded as early as they are is because I want to give people as much opportunity as possible to do whatever they want to do and having C++ mods load as early as possible means they could potentially do things before UE inits like establish a proper lifetime cycle. Any delays make this more difficult as you decrease the time available for a mod to init before UE inits. Not sure how much time the output configuration takes but it's probably a low amount so your change is likely fine.
> 
> Ah this makes sense. I'm not aware of how long the GUI takes to init, but if it's a problem, new tabs could be queued by `UE4SSProgram` so they don't have to wait. This brings up something I forgot to mention last night: Does the GUI ever not exist, or does it just hide when disabled? So far I haven't considered what happens if it's disabled in the config.

The GUI isn't initialized at all if `GuiConsoleEnabled` is set to 0 in UE4SS-settings.ini.
If that setting is set to 1, and `GuiConsoleVisible` is set to 0, the GUI is not initialized but the logging system is enabled for it, the UObjectArray listeners are set up and the hotkey for opening the GUI is created and when that hotkey is hit the GUI will be initialized.

> > You're using make_unique and then assigning the result to a shared_ptr. I'm guessing this was a simple mistake but I thought it was worth mentioning anyway.
> 
> It was my understanding that uniqe_ptr will cheaply and automatically get converted to another pointer type when necessary, so haven't been bothering to use anything else when creating a pointer. Is this not the case? Either way I guess it doesn't hurt to be explicit.

While it's technically fine to use `unique_ptr` in this case, it's also misleading because it looks like the variable is a `unique_ptr` simply because you're creating a `unique_ptr`.

PR #115, opened 2023-05-08
Build number cleanup

trumank:
These changes stem primarily from running cmake causing `generated_src/verison.cache` to appear dirty to git.

I also looked at removing `generated_include/MacroSetter.hpp` as it doesn't appear to be actually used anywhere. Is this the case?


UE4SS:
The file `MacroSetter.hpp` is used in `UE4SSProgram.cpp` on line 364.

trumank:
I was unclear, sorry. I meant I don't see anything ever writing to `MacroSetter.hpp` in the cmake files or otherwise.

UE4SS:
The generator that creates `MacroSetter.hpp` and also `Unreal/generated_include` is not included in this repo.

PR #122, opened 2023-05-17
Initial generators+views implementation

localcc:
This patch removes the need for passing lambdas into `ForEach` functions in a lot of places, instead replacing the iterator producing functions with coroutines that yield the results.

This change allows for more effective work with any iterable by allowing to construct STL containers from them, use ranges on the iterables, etc.

Coroutines also allow for much faster iteration than the current method when compiled with clang.


localcc:
CI's not gonna pass until Unreal submodule gets merged

UE4SS:
Thanks for your contribution.

PR #124, opened 2023-05-23
Log UE modules

trumank:
Untested against any modular games

### DIFF
--- a/src/UE4SSProgram.cpp
+++ b/src/UE4SSProgram.cpp
@@ -214,6 +214,20 @@ namespace RC
             Output::send(STR("object dumper directory: {}\n\n\n"), m_object_dumper_output_directory.c_str());
 
             setup_unreal();
+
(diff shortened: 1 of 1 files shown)


narknon:
I'd prefer this go after the unreal init and unreal properties calls, if possible.

trumank:
> I'd prefer this go after the unreal init and unreal properties calls, if possible.
What's your reasoning behind this? Logging earlier provides more context in case something does fail, and the next line is `fire_unreal_init_for_cpp_mods()` which is third party mods, so a likely place to fail.

narknon:
That's true, I'm just hesitant to slow down the init whenever it can be avoided and I'm not sure that calculating the size of the module is so imperative for debugging that it is worth the extra processing that early.

trumank:
It's only logging structs that have been populated in a previous stage of initialization. The most time consuming part is probably formatting and logging to the console, which are minimal. I understand the concern for time-to-init, and in this case think it's not an issue. Time-to-init would be a good benchmark to set up so this isn't just speculation.

PR #127, opened 2023-05-24
Fix case preserving names switch

localcc:
Also make the switch more idiomatic with CMake options instead of reading files.


UE4SS:
Thanks for this.
It appears at some point this define was removed from CMakeLists and it's good to have it back again.

PR #137, opened 2023-06-28
Update uscriptstruct with examples of index and newindex usage

Buckminsterfullerene02:
The current explanation is misleading if you don't have any knowledge of the concept of Lua metafunctions. Hopefully I have explained this correctly with valid examples.


Buckminsterfullerene02:
Better?

UE4SS (approved):
This looks good.

PR #144, opened 2023-07-09
fixes for various unload & executeingamethread gc-related crashes

noita-player:
current behavior is to infinitely grow the mod's stack until we write out of bounds on the heap.

note that `mod->lua().get_lua_state()`, `hook_lua->get_lua_state()` and `lua.get_lua_state()` are three different states.

verified no crashes on reload now with gflags +hpa and LUA_USE_APICHECK enabled

### DIFF
--- a/include/Mod/LuaMod.hpp
+++ b/include/Mod/LuaMod.hpp
@@ -46,6 +46,7 @@ namespace RC
         {
             const LuaMadeSimple::Lua* lua;
             int32_t lua_action_function_ref{};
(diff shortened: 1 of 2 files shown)


UE4SS:
Thank you!

UE4SS:
I accepted this pull request too quickly.
There is a problem on line 2622 in LuaMod.cpp
See this: https://github.com/UE4SS-RE/RE-UE4SS/actions/runs/5507696841/jobs/10037992881
Could you fix this and resubmit this pull request ?

PR #145, opened 2023-07-10
fix build error introduced by #144

noita-player:
my apologies, I'm merging these changes down from a ver with a lot more changes that aren't ready for PR and missed this line.

this does build clean locally, but I left that line staged for commit.

### DIFF
--- a/include/Mod/LuaMod.hpp
+++ b/include/Mod/LuaMod.hpp
@@ -46,6 +46,7 @@ namespace RC
         {
             const LuaMadeSimple::Lua* lua;
             int32_t lua_action_function_ref{};
+            int32_t lua_action_thread_ref{};
         };
 
         struct AsyncAction
--- a/src/Mod/LuaMod.cpp
+++ b/src/Mod/LuaMod.cpp
@@ -2558,6 +2558,9 @@ Overloads:
 
             TRY([&]() { lua_data.lua->call_function(0, 0); });
 
+            // thread_ref came from lua_newthread, we can let it GC now.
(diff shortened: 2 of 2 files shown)


noita-player:
yep, the intention would be to allow it to GC at some point, but I haven't tested unrefing it in UnregisterHook (as my test cases just don't unhook)

current behavior will be to leak a registryindex entry on every UnregisterHook call, which is what is done for lua_callback_ref as well. a suite of tests that work on shootergame would be nice, like a test that would expose this registry leak would be registerhook/unregisterhook N times and memory usage would be inflated by something like N*8 bytes (or crash when we run out of registry space)

noita-player:
implemented unrefing, tested with pageheap and LUA_USE_APICHECK 
```lua
RegisterKeyBind(Key.A, {ModifierKey.CONTROL}, function()
    print("[Main] RegisterHook test...")
    funcName = "/Script/Engine.PlayerController:ClientRestart"
    for i=1,1000 do
        local a,b = RegisterHook(funcName, function() end)
        UnregisterHook(funcName, a,b)
    end
        print("[EIGT] RegisterHook test...")
        funcName = "/Script/Engine.PlayerController:ClientRestart"
        for i=1,1000 do
            local a,b = RegisterHook(funcName, function() end)
            UnregisterHook(funcName, a,b)
        end
    end)
    LoopAsync(2000, function()
        print("[Async] RegisterHook test...")
        funcName = "/Script/Engine.PlayerController:ClientRestart"
        for i=1,1000 do
            local a,b = RegisterHook(funcName, function() end)
            UnregisterHook(funcName, a,b)
        end
    end)
end)
```

UE4SS:
Thank you!

PR #177, opened 2023-08-05
Added Mod Testing Automation Scripts

Mythical-Github:
Added some mod testing automation scripts for easy C++ dll modding setup. Essentially to use, you bring the bat file, up into your MyMods folder, and edit the build_and_testing.py, for your game then running the bat should handle the rest
I've tested it with several games
Remnant 2, Amid Evil, and Lies of P Demo

### DIFF
--- a/tools/buildscripts/internal_build_tools/generate_vs_solution_internal_debug.bat
+++ b/tools/buildscripts/internal_build_tools/generate_vs_solution_internal_debug.bat
@@ -1,0 +1,8 @@
+set BuildMode=%1
+
+IF NOT DEFINED %BuildMode (
+set BuildMode=Debug
+)
+
+cd ../../../VS_Solution
+cmake -DCMAKE_BUILD_TYPE=%BuildMode% -DRC_FORCE_ALL_STATIC_LIBS= -G"Visual Studio 17 2022" ..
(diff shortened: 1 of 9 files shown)


UE4SS:
I'm not sure this is needed.
It seems it just builds the mod and copies the files to a hard-coded path.

Building the mod shouldn't be a problem to do manually in an IDE or using CLI, I don't think we need yet another script for that.

Copying the files might be a good idea but it might be better as a post-build command in CMake so that it happens automatically when the mod is built.
In any case, there has been talk of a mod template and if such a thing is ever made, this should probably go in there instead.

Here's a list of issues I have with the scripts:
1. I don't like how much hard-coding of paths there is. At least, MSBuild shouldn't need to be hard-coded since we're dependent on it being in PATH for other scripts.
2. There's no reason to create debug/release specific versions of `generate_vs_solution_internal.bat`, you should pass `Debug` or `Release` to it instead.
3. The PDB files do not need to be copied to the game directory. Symbols should be loaded from the build directory without problems.

UE4SS:
Furthermore, when it comes to supplying the path to the game directory, instead of having a build script that every dev of a mod must modify, I think it's better to have a separate config file that stores the path and this file can be untracked to avoid making a mess when committing.
Having a separate file would also avoid conflicts in cases where the build script is updated for other reasons.

Mythical-Github:
Thanks for the in depth explanation
I'll try and polish this up in the coming days

PR #187, opened 2023-08-26
UHT Generator updates for component creation and attachment, as well as property overrides for inherited properties and objectinitializer overrides

narknon:
* Allow inherited property initializers to be generated if they are different from the parent class' CDO.
* Add objectinitializer to all AActor child classes
* Set root components
* Set Component class overrides
* Set component attachment


narknon:
Replaced with PR #286

PR #193, opened 2023-08-30
Trigger website update on release

localcc:


UE4SS:
Is this dead or relevant at all ? We just had a release without this PR being merged, and I didn't notice any problems.

Buckminsterfullerene02:
> Is this dead or relevant at all ? We just had a release without this PR being merged, and I didn't notice any problems.

No, I think the idea is to move the UE4SS website (seperate to the docs site) to a seperate repo but it hasn't been done yet.

PR #201, opened 2023-09-16
Add SCP: 5K signatures

mitchell-merry:
As the readme suggests, this game is in early access and updates often, so I imagine these may break at some point.

Thanks for the docs, I found them helpful.


UE4SS:
Thank you.

PR #234, opened 2023-10-19
added crash handler

localcc:
This crash handler allows for crashdump generation.

### DIFF
--- a/assets/UE4SS-settings.ini
+++ b/assets/UE4SS-settings.ini
@@ -133,6 +133,10 @@ HookBeginPlay  = 1
 HookLocalPlayerExec = 1
 FExecVTableOffsetInLocalPlayer = 0x28
 
+[CrashDump]
+EnableDumping = 1
+FullMemoryDump = 0
+
 [ExperimentalFeatures]
 ; Only enable these features if you know what you are doing.
 GUIUFunctionCaller = 1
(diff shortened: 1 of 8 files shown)


assets/UE4SS-settings.ini
```diff
@@ -87,6 +87,7 @@ MakeAllConfigsEngineConfig = 1
 ConsoleEnabled = 1
 GuiConsoleEnabled = 1
 GuiConsoleVisible = 1
+FullMemoryDump = 0
```
narknon, line 90: I don't really like this being grouped.  At this point the live view debugger is more of a mod debugger than a UE4SS debugger, and I think the 2 concepts should be distinguished.

narknon:
Please add an option to disable it completely (but maybe move the settings all the way to the bottom above experimental features).

UE4SS:
> Please add an option to disable it completely (but maybe move the settings all the way to the bottom above experimental features).

What is the purpose of this ?

narknon:
Would you want a program that dumped 20mb of data every time it crashed with absolutely no option to turn it off? Hell, I don't want it on when I'm testing stuff I know will probably crash 50 times.

UE4SS:
> Would you want a program that dumped 20mb of data every time it crashed with absolutely no option to turn it off? Hell, I don't want it on when I'm testing stuff I know will probably crash 50 times.

Personally, wouldn't affect me, I'm always gonna have it enabled, but I see your point.

narknon:
Tfw you forget your c drive is almost full and end up zip bombed by a crash dump (UE has done this to me several times)

PR #251, opened 2023-11-01
Made the CLion fixes optional

UE4SS:
These fixes were the project to not link if actually built with CLion.


narknon:
So with this inactive, CLion will build but it will have a ton of linter errors; and with it active the linter errors are gone but CLion won't build?

UE4SS:
> So with this inactive, CLion will build but it will have a ton of linter errors; and with it active the linter errors are gone but CLion won't build?

Yes, that has been my experience.

narknon:
Ok, I think we merge this for now but create an issue so that we can look for better resolutions at some point.  Do you mind creating the issue?

UE4SS:
Issue made here: #252

PR #280, opened 2023-11-06
feat: Added LUA bindings for enum mutation functions

CpazR:
Added LUA bindings for the following UEnum funcitons:
- GetEnumNameByIndex
- InsertIntoNames
- EditNameAt
- EditValueAt
- RemoveFromNamesAt

DRAFT: Implementations are still POC. There's also likely more documentation that could be added, alongside actual LUA helpers.

### DIFF
--- a/UE4SS/src/LuaType/LuaUEnum.cpp
+++ b/UE4SS/src/LuaType/LuaUEnum.cpp
@@ -175,7 +175,7 @@ Overloads:
             }
             else
             {
-                lua.throw_error("'UEnum.InsertIntoNames' could not load parameter for \"Index\"");
+                lua.throw_error("'UEnum.InsertIntoNames' could not load parameter for \"Value\"");
             }
 
             // P3 (Index), integer
@@ -319,17 +319,17 @@ Overloads:
             bool param_allow_shrinking = true;
 
             // P1 (Index), integer
-            if (lua.is_string())
+            if (lua.is_integer())
             {
                 param_index = lua.get_integer();
             }
             else
             {
-                lua.throw_error("'UEnum.RemoveFromNamesAt' Could not load parameter for \"Name\"");
+                lua.throw_error("'UEnum.RemoveFromNamesAt' Could not load parameter for \"Index\"");
             }
 
             // P2 (Count), integer
-            if (lua.is_string())
+            if (lua.is_integer())
             {
                 param_count = lua.get_integer();
             }
--- a/UE4SS/src/LuaType/LuaUEnum.cpp
+++ b/UE4SS/src/LuaType/LuaUEnum.cpp
@@ -304,5 +304,17 @@ Overloads:
             lua_object.get_remote_cpp_object()->RemoveFromNamesAt(param_index);
             return 1;
         });
+
+        if constexpr (is_final == LuaMadeSimple::Type::IsFinal::Yes)
+        {
+            table.add_pair("type", [](const LuaMadeSimple::Lua& lua) -> int {
+                lua.set_string(ClassName::ToString());
+                return 1;
+            });
+
+            // If this is the final object then we also want to finalize creating the table
+            // If not then it's the responsibility of the overriding object to call 'make_global()'
+            // table.make_global(ClassName::ToString());
+        }
     }
 } // namespace RC::LuaType
--- a/UE4SS/src/LuaType/LuaUEnum.cpp
+++ b/UE4SS/src/LuaType/LuaUEnum.cpp
@@ -233,7 +233,7 @@ Overloads:
             }
 
             // P2 (NewName), string
-            if (lua.is_integer())
+            if (lua.is_string())
             {
                 param_new_name = lua.get_string();
             }
--- a/UE4SS/src/LuaType/LuaUEnum.cpp
+++ b/UE4SS/src/LuaType/LuaUEnum.cpp
@@ -110,16 +110,199 @@ Overloads:
             return 0;
         });
 
-        if constexpr (is_final == LuaMadeSimple::Type::IsFinal::Yes)
-        {
-            table.add_pair("type", [](const LuaMadeSimple::Lua& lua) -> int {
-                lua.set_string(ClassName::ToString());
-                return 1;
-            });
+        table.add_pair("GetEnumNameByIndex", [](const LuaMadeSimple::Lua& lua) -> int {
+            std::string error_overload_not_found{R"(
+No overload found for function 'UEnum.InsertIntoNames'.
+Overloads:
+#1: GetNameByValue(integer Value))"};
 
-            // If this is the final object then we also want to finalize creating the table
-            // If not then it's the responsibility of the overriding object to call 'make_global()'
-            // table.make_global(ClassName::ToString());
-        }
+            auto& lua_object = lua.get_userdata<UEnum>();
+
+            if (!lua.is_integer())
+            {
+                lua.throw_error(error_overload_not_found);
+            }
+
+            auto value = lua.get_integer();
+            LuaType::FName::construct(lua, lua_object.get_remote_cpp_object()->GetEnumNameByIndex(value));
+            return 1;
+        });
+
+        table.add_pair("InsertIntoNames", [](const LuaMadeSimple::Lua& lua) -> int {
+            // Function: InsertIntoNames
+            // Param #1: Enum key name
+            // Param #2: Enum key index
+            std::string error_overload_not_found{R"(
+No overload found for function 'UEnum.InsertIntoNames'.
+Overloads:
+#1: InsertIntoNames(string Name, integer Index) // shiftValues = true
+#1: InsertIntoNames(string Name, integer Index, bool shiftValues)"};
+
+            int32_t stack_size = lua.get_stack_size();
+
+            if (stack_size <= 0)
+            {
+                lua.throw_error("Function 'UEnum.InsertIntoNames' cannot be called with 0 parameters.");
+            }
+
+            StringType param_name = L"";
+            int param_index = 0;
+            bool param_shift = true;
+
+            // P1 (Name), string
+            if (lua.is_string())
+            {
+                param_name = to_wstring(lua.get_string());
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"Name\"");
+            }
+
+            // P2 (Index), integer
+            if (lua.is_integer())
+            {
+                param_index = lua.get_integer();
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"Index\"");
+            }
+
+            // (Optional) P3 (ShiftValues), boolean
+            if (lua.is_bool())
+            {
+                param_index = lua.get_bool();
+            }
+
+            auto& lua_object = lua.get_userdata<UEnum>();
+
+            Unreal::FName key = Unreal::FName(param_name, Unreal::FNAME_Add);
+            auto pair = Unreal::TPair{key, Unreal::int64(param_index)};
+
+            lua_object.get_remote_cpp_object()->InsertIntoNames(pair, param_index, true);
+            return 1;
+        });
+
+        table.add_pair("EditNameAt", [](const LuaMadeSimple::Lua& lua) -> int {
+            // Function: EditNameAt
+            // Param #1: Enum key index
+            // Param #2: New name for index
+            std::string error_overload_not_found{R"(
+No overload found for function 'UEnum.EditNameAt'.
+Overloads:
+#1: EditNameAt(integer Index, string NewName))"};
+
+            int32_t stack_size = lua.get_stack_size();
+
+            if (stack_size <= 0)
+            {
+                lua.throw_error("Function 'UEnum.EditValueAt' cannot be called with 0 parameters.");
+            }
+
+            int param_index = 0;
+            std::string param_new_name = L"";
+
+            // P1 (Index), integer
+            if (lua.is_integer())
+            {
+                param_index = lua.get_integer();
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"Index\"");
+            }
+
+            // P2 (NewName), string
+            if (lua.is_integer())
+            {
+                param_new_name = to_wstring(lua.get_string());
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"NewName\"");
+            }
+
+            auto& lua_object = lua.get_userdata<UEnum>();
+            Unreal::FName new_key = Unreal::FName(param_new_name, Unreal::FNAME_Add);
+            lua_object.get_remote_cpp_object()->EditNameAt(param_index, new_key);
+            return 1;
+        });
+
+        table.add_pair("EditValueAt", [](const LuaMadeSimple::Lua& lua) -> int {
+            // Function: EditValueAt
+            // Param #1: Enum key Index
+            // Param #2: New value for Index
+            std::string error_overload_not_found{R"(
+No overload found for function 'UEnum.EditValueAt'.
+Overloads:
+#1: EditValueAt(integer Index, integer NewValue))"};
+
+            int32_t stack_size = lua.get_stack_size();
+
+            if (stack_size <= 0)
+            {
+                lua.throw_error("Function 'UEnum.EditValueAt' cannot be called with 0 parameters.");
+            }
+
+            int param_index = 0;
+            int param_new_value = 0;
+
+            // P1 (Index), integer
+            if (lua.is_integer())
+            {
+                param_index = lua.get_integer();
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"Index\"");
+            }
+
+            // P2 (NewValue), integer
+            if (lua.is_integer())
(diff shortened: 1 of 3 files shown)


UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+
+            auto& lua_object = lua.get_userdata<UEnum>();
+
+            if (!lua.is_integer())
+            {
+                lua.throw_error(error_overload_not_found);
+            }
+
+            auto value = lua.get_integer();
+            LuaType::FName::construct(lua, lua_object.get_remote_cpp_object()->GetEnumNameByIndex(value));
```
UE4SS, line 127: This should ideally mimic `UEnum::GetEnumNameByIndex` by returning both the name and the value.
We could make use of multiple return values that Lua supports here by constructing both values and `return 2` instead of `return 1`.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
-                return 1;
-            });
-
-            // If this is the final object then we also want to finalize creating the table
-            // If not then it's the responsibility of the overriding object to call 'make_global()'
-            // table.make_global(ClassName::ToString());
-        }
+        table.add_pair("GetEnumNameByIndex", [](const LuaMadeSimple::Lua& lua) -> int {
+            std::string error_overload_not_found{R"(
+No overload found for function 'UEnum.InsertIntoNames'.
```
UE4SS, line 115: Should be `GetEnumNameByIndex`, not `InsertIntoNames`.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+            {
+                lua.throw_error(error_overload_not_found);
+            }
+
+            auto value = lua.get_integer();
+            LuaType::FName::construct(lua, lua_object.get_remote_cpp_object()->GetEnumNameByIndex(value));
+            return 1;
+        });
+
+        table.add_pair("InsertIntoNames", [](const LuaMadeSimple::Lua& lua) -> int {
```
UE4SS, line 131: This is not how `InsertIntoNames` works.
The first param is the FName/int64 pair, the second param (the 'index') is the index into the array where this new pair will be inserted.
In Lua, since we don't have a pair, we have a couple of choices:
1. Use an additional param, making the signature: `InsertIntoNames(string Name, integer Value, integer Index, bool bShiftValues)`.
2. Use a Lua table to represent the Name/Value pair.

I have not reviewed the rest of this function because it needs to be reworked.
Will review this function after the rework.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+            std::string param_new_name = L"";
+
+            // P1 (Index), integer
+            if (lua.is_integer())
+            {
+                param_index = lua.get_integer();
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"Index\"");
```
UE4SS, line 213: This error message is not descriptive enough.
Add which function is throwing the error.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+            }
+
+            // P2 (NewName), string
+            if (lua.is_integer())
+            {
+                param_new_name = to_wstring(lua.get_string());
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"NewName\"");
```
UE4SS, line 223: This error message is not descriptive enough.
Add which function is throwing the error.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+
+            int32_t stack_size = lua.get_stack_size();
+
+            if (stack_size <= 0)
+            {
+                lua.throw_error("Function 'UEnum.EditValueAt' cannot be called with 0 parameters.");
+            }
+
+            int param_index = 0;
+            std::string param_new_name = L"";
```
UE4SS, line 204:
```suggestion
            std::string param_new_name{};
```
There's no reason to construct an empty string here, use the default constructor instead.
Also, do not use `L` to create strings, always use the `STR("string here")` macro instead.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+            if (lua.is_integer())
+            {
+                param_new_name = to_wstring(lua.get_string());
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"NewName\"");
+            }
+
+            auto& lua_object = lua.get_userdata<UEnum>();
```
UE4SS, line 226: This probably causes a crash, and if not, at least the parameter values won't be correct.
This should be the first thing retrieved from Lua, move this to the top of the function.
We always retrieve parameters (including the implicit `this`/`context`/`self` param) in the correct order.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+            int param_new_value = 0;
+
+            // P1 (Index), integer
+            if (lua.is_integer())
+            {
+                param_index = lua.get_integer();
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"Index\"");
```
UE4SS, line 258: This error message is not descriptive enough.
Add which function is throwing the error.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+            }
+
+            // P2 (NewValue), integer
+            if (lua.is_integer())
+            {
+                param_new_value = lua.get_integer();
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"NewValue\"");
```
UE4SS, line 268: This error message is not descriptive enough.
Add which function is throwing the error.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+            }
+
+            auto& lua_object = lua.get_userdata<UEnum>();
```
UE4SS, line 271: This probably causes a crash, and if not, at least the parameter values won't be correct.
This should be the first thing retrieved from Lua, move this to the top of the function.
We always retrieve parameters (including the implicit this/context/self param) in the correct order.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+            int param_index = 0;
+
+            // P1 (Index), integer
+            if (lua.is_string())
+            {
+                param_index = lua.get_integer();
+            }
+            else
+            {
+                lua.throw_error("ERROR: Could not load parameter for \"Name\"");
```
UE4SS, line 300: This error message is not descriptive enough.
Add which function is throwing the error.

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,16 +110,199 @@ No overload found for function 'UEnum.ForEachName'.
...
+            }
+
+            auto& lua_object = lua.get_userdata<UEnum>();
```
UE4SS, line 303: This probably causes a crash, and if not, at least the parameter values won't be correct.
This should be the first thing retrieved from Lua, move this to the top of the function.
We always retrieve parameters (including the implicit this/context/self param) in the correct order.

UE4SS (changes requested):
On top of what I mentioned in this review, all params that are taking a Lua string and constructing an FName with it should also be able to take an FName already constructed in Lua.

UE4SS/src/LuaType/LuaUEnum.cpp
CpazR: I believe this is setup correctly now. Please verify to see if it makes sense.

UE4SS/src/LuaType/LuaUEnum.cpp
CpazR: Reworked this. Let me know if there's any issues.

UE4SS/src/LuaType/LuaUEnum.cpp
CpazR: Updated error messages for clarity.

UE4SS/src/LuaType/LuaUEnum.cpp
CpazR: Interesting. First time working with lua like this. Is this a general restriction of implementing lua, or is this related to this specific implementation?

UE4SS/src/LuaType/LuaUEnum.cpp
UE4SS: Looks good.

docs/lua-api/classes/uenum.md
CpazR, line 1: General question: Is it possible to add helpers to these functions or is the documentation good enough?

Based on what I saw in the shared helpers, I assume not.

UE4SS/src/LuaType/LuaUEnum.cpp
UE4SS: The Lua VM uses a stack for values and when a function is called from Lua, the params are pushed onto the stack in a specific order.
Usually with Lua, you need manually specify which index on the stack that you want to access but with our API you don't need to do that most of the time as long as you take things from the stack in the correct order.

docs/lua-api/classes/uenum.md
UE4SS: > General question: Is it possible to add helpers to these functions or is the documentation good enough?
> 
> Based on what I saw in the shared helpers, I assume not.

What kind of helpers ?

docs/lua-api/classes/uenum.md
```diff
@@ -14,4 +14,23 @@
 
 - Iterates every `FName`/`Value` combination that belongs to this enum.
 - The callback has two params: `FName Name`, `integer Value`.
-- Return `true` in the callback to stop iterating.
\ No newline at end of file
+- Return `true` in the callback to stop iterating.
+
+### GetEnumNameByIndex(integer Index)
+- **Return type:** `FName`/`Integer`
```
UE4SS, line 20:
```suggestion
- **Return types:** `FName`, `Integer`
```
A more clear way to put this is: `Return types: FName, Integer`

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,6 +110,240 @@ No overload found for function 'UEnum.ForEachName'.
...
+        table.add_pair("InsertIntoNames", [](const LuaMadeSimple::Lua& lua) -> int {
+            // Function: InsertIntoNames
+            // Param #1: Enum key name
+            // Param #2: Enum value
+            // Param #3: Enum index
+            // (Optional) Param #4: Shift values
+            std::string error_overload_not_found{R"(
+No overload found for function 'UEnum.InsertIntoNames'.
+Overloads:
+#1: InsertIntoNames(string Name, integer Value, integer Index) // shiftValues = true
```
UE4SS, line 144: The default in UEnum.hpp is actually `false`, not `true`.
Why does the Lua binding deviate from this ?

CpazR: Double checked and you're right. Dunno how I missed that. Fixed.

docs/lua-api/classes/uenum.md
CpazR: I think I'm referring to what visual studio code uses for intellisense on some functions (I think it's in under `Mods/shared/Types.lua` or something like that)
But it looked like was only for global functions.

UE4SS: > I think I'm referring to what visual studio code uses for intellisense on some functions (I think it's in under `Mods/shared/Types.lua` or something like that) But it looked like was only for global functions.

The file is not just for global functions, it is however incomplete and is missing the entire UEnum class.
You should be able to add it as long as you keep to the exact same format as all other classes.

docs/lua-api/classes/uenum.md
```diff
@@ -12,6 +12,25 @@
 
 ### ForEachName(LuaFunction Callback)
 
-- Iterates every `FName`/`Value` combination that belongs to this enum.
+- Iterates every `FName`, `Value` combination that belongs to this enum.
```
UE4SS, line 15:
```suggestion
- Iterates every `FName`/`Value` combination that belongs to this enum.
```
The `/` was appropriate here and is more clean than a `,` for this usage of the pair.

docs/lua-api/classes/uenum.md
```diff
@@ -12,6 +12,25 @@
...
 - The callback has two params: `FName Name`, `integer Value`.
-- Return `true` in the callback to stop iterating.
\ No newline at end of file
+- Return `true` in the callback to stop iterating.
+
+### GetEnumNameByIndex(integer Index)
+- **Return type:** `FName`, `Integer`
```
UE4SS, line 20:
```suggestion
- **Return types:** `FName`, `Integer`
```
Grammatical error. I want it to be very clear that this is multiple return values.

docs/lua-api/classes/uenum.md
```diff
@@ -12,6 +12,25 @@
...
+- **Returns:** the `FName` that coresponds the given `Index`.
+- **Returns:** the `Integer` value that coresponds the given `Index`.
+
+### InsertIntoNames(string Name, integer Value, integer Index, boolean ShiftValues = true)
+- Inserts a `Name/Value` combination into a a `UEnum` at the given `Index`.
```
UE4SS, line 25:
```suggestion
- Inserts a `FName`/`Value` combination into a a `UEnum` at the given `Index`.
```
Consistency with how pairs are displayed in this file.

UE4SS (changes requested):

docs/lua-api/classes/uenum.md
```diff
@@ -14,4 +14,23 @@
...
+- **Return types:** `FName`, `Integer`
+- **Returns:** the `FName` that coresponds the given `Index`.
+- **Returns:** the `Integer` value that coresponds the given `Index`.
+
+### InsertIntoNames(string Name, integer Value, integer Index, boolean ShiftValues = true)
+- Inserts a `FName`/`Value` combination into a a `UEnum` at the given `Index`.
+- If `ShiftValues = true`, will shift all enum values greater than inserted value by one.
+
+### EditNameAt(integer Index, string NewName)
+- At a given `Index`, will modify the found element in the `UEnum` and replace it's `Name` with the given `NewName`.
```
narknon, line 29: Nitpicky, but can you fix the `it's` that should be `its`?

assets/Mods/shared/Types.lua
```diff
@@ -819,4 +819,66 @@ function RemoteUnrealParam:type() end
...
+---@class TSubclassOf<T> : UClass
+
+
+---@class UEnum
+
+UEunm = {}
+
+--- Returns the `FName` that corresponds to the specified value.
+---@param Value integer
+function UEunm:GetNameByValue(Value) end
```
UE4SS, line 831: All the instances of `UEnum` are typo'd as `UEunm` in this file.

CpazR: Fixed.

UE4SS (approved):

UE4SS/src/LuaType/LuaUEnum.cpp
```diff
@@ -110,6 +110,240 @@ No overload found for function 'UEnum.ForEachName'.
...
+            // (Optional) P4 (ShiftValues), boolean
+            if (lua.is_bool())
+            {
+                param_shift = lua.get_bool();
+            }
+
+            Unreal::FName key = Unreal::FName(param_name, Unreal::FNAME_Add);
+            auto pair = Unreal::TPair{key, Unreal::int64(param_index)};
+
+            lua_object.get_remote_cpp_object()->InsertIntoNames(pair, param_index, true);
```
narknon, line 200: This `true` should be `param_shift` I believe

narknon (commented):
It doesn't have to make it into this PR, but we should have an option to return the enum array so that it is easier to get the Num()/treat it as an array without foreaching.

UE4SS/src/LuaType/LuaUEnum.cpp
CpazR: No clue how I keep missing stuff like this...
Fixed.

PR #299, opened 2024-01-09
fix: add a basic Lua mod creation tutorial

massclown:
Add a basic tutorial how to create a Lua mod in UE4SS. 
The included example should be basic and generic enough to run on a number of supported games.


narknon:
This looks good, thank you!

PR #301, opened 2024-01-14
fix: add missing FText Lua bindings and a global FText(string) function

massclown:
* Add metamethods for the Lua bindings of FText class allowing to construct it
* Add global function FText(string) and a table during mod initialization
* Add initial documentation for Lua global function

Most things were implemented in a dumb manner, by looking at FName implementation.


UE4SS/src/LuaType/LuaFText.cpp
```diff
@@ -59,6 +98,18 @@ namespace RC::LuaType
             return 1;
         });
 
+        table.add_pair("Equals", [](const LuaMadeSimple::Lua& lua) -> int {
```
UE4SS, line 101: I don't know if it's a good idea to have an `Equals` function for `FText`.
It exists for `FName` because it actually exists in UE as well and does something different than `operator==`/`__eq`.

I haven't looked into whether something similar to `Equals` exists in UE for `FText`, but if it does, that behavior should probably be mimicked here if possible, and if it's not possible, we shouldn't define this function until it's possible to mimic the exact behavior.

UE4SS/src/LuaType/LuaFText.cpp
```diff
@@ -42,9 +42,48 @@ namespace RC::LuaType
...
+        base_object.get_metamethods().create(LuaMadeSimple::Lua::MetaMethod::Equal, []([[maybe_unused]] const LuaMadeSimple::Lua& lua) -> int {
+            if (!lua.is_userdata(1) || !lua.is_userdata(2))
+            {
+                lua.throw_error("FText __eq metamethod called but there was not two userdata to compare");
+            }
+
+            auto name_a = lua.get_userdata<LuaType::FText>();
+            auto name_b = lua.get_userdata<LuaType::FText>();
+            // FText objects are equal if their string representations are equal
+            return name_a.get_local_cpp_object().ToString() == name_b.get_local_cpp_object().ToString();
```
UE4SS, line 85:
```suggestion
            auto text_a = lua.get_userdata<LuaType::FText>();
            auto text_b = lua.get_userdata<LuaType::FText>();
            // FText objects are equal if their string representations are equal
            return text_a.get_local_cpp_object().ToString() == text_b.get_local_cpp_object().ToString();
```

UE4SS/src/LuaType/LuaFText.cpp
```diff
@@ -59,6 +98,18 @@ namespace RC::LuaType
...
+        table.add_pair("Equals", [](const LuaMadeSimple::Lua& lua) -> int {
+            if (!lua.is_userdata(1) || !lua.is_userdata(2))
+            {
+                lua.throw_error("FText.Equals called but there was not two userdata to compare (use ':' to call, not '.')");
+            }
+
+            auto name_a = lua.get_userdata<LuaType::FText>();
+            auto name_b = lua.get_userdata<LuaType::FText>();
+
+            return name_a.get_local_cpp_object().ToString() == name_b.get_local_cpp_object().ToString();
```
UE4SS, line 110:
```suggestion
            auto text_a = lua.get_userdata<LuaType::FText>();
            auto text_b = lua.get_userdata<LuaType::FText>();

            return text_a.get_local_cpp_object().ToString() == text_b.get_local_cpp_object().ToString();
```

UE4SS/src/LuaType/LuaFText.cpp
massclown: Thanks, will look it up and either fix or remove.

massclown: There was no "Equals" in UE sources, so removed this one. There is only an `EqualsTo` with an additional `ComparisonLevel` parameter (they explain that localized strings cannot be compared simply & directly), which is not anywhere in the headers we use, so I would rather not implement that logic here unless the users request it for something specific.

ViniciusFXavier (approved):

Buckminsterfullerene02:
@UE4SS any last thoughts before merge?

UE4SS:
> @UE4SS any last thoughts before merge?

It looks good to me.

PR #304, opened 2024-01-23
Replace built-in UE init scans with patternsleuth

trumank:
This pulls in patternsleuth and uses its Unreal Engine resolvers which should have the same or better game coverage than what currently exists in RE-UE4SS. I have yet to build string based resolvers for every symbol but for the ones I have, they are nearing 100% coverage over our test game collection. Using patternsleuth also makes measuring coverage and improvements possible because all the scans are run offline. Coverage as of this commit:

```
+===============================+=========+
| EngineVersion                 | 98.19%  |
+-------------------------------+---------+
| FNameCtorWchar                | 98.68%  |
+-------------------------------+---------+
| FNameToString                 | 99.31%  |
+-------------------------------+---------+
| FTextFString                  | 92.08%  |
+-------------------------------+---------+
| GMalloc                       | 100.00% |
+-------------------------------+---------+
| GUObjectArray                 | 94.03%  |
+-------------------------------+---------+
| StaticConstructObjectInternal | 99.03%  |
+-------------------------------+---------+

```


UE4SS:
What's your reasoning for pulling in patternsleuth as a submodule ?
Are there any problem with using CMake like we do with the other dependencies ?

UE4SS:
We should add the minimum required version of rustc to the readme.
I tried with 1.70 and that was too old, not sure where the breaking point is but 1.75 works.
This built for me both in my VM and using Proton on Linux out of the box after updating rust.

UE4SS:
For future reference, these are the steps I had to take to get this to build:

On Windows, in a VM on a Linux host, I installed rustup, and then ran CMake and it built without any problem.

On Linux with Wine (using https://github.com/mstorsjo/msvc-wine), I first had to update my rust installation from 1.70 to 1.75, and then I had to add the toolchain to rust as prompted by the error message, and then I had to delete the build directory and re-run CMake and after that it built without problems.

On both, since this isn't merged yet, I also had to go to `deps/first/Unreal` and execute `git checkout patternsleuth`.

trumank:
> What's your reasoning for pulling in patternsleuth as a submodule ? Are there any problems with using CMake like we do with the other dependencies ?

That's the way I originally had it but it makes iterating on patternsleuth difficult as each change has to be pushed to a remote repo or temporarily pointed to a local directory while testing.

UE4SS:
> > What's your reasoning for pulling in patternsleuth as a submodule ? Are there any problems with using CMake like we do with the other dependencies ?
> 
> That's the way I originally had it but it makes iterating on patternsleuth difficult as each change has to be pushed to a remote repo or temporarily pointed to a local directory while testing.

That's unfortunate, but if you're using UE4SS as the main test bed then I guess it does make more sense to use submodules for actively developed out-of-repo first-party dependencies than CMake's FetchContent.

Buckminsterfullerene02:
Tested on a bunch of my games with small lua mods, everything seems to work fine. 

We good to merge?

UE4SS:
> Tested on a bunch of my games with small lua mods, everything seems to work fine.
> 
> We good to merge?

I have no objections to this.

PR #306, opened 2024-01-23
Lua hooks (RegisterHook) can now access the original return value

UE4SS:
The second function passed to 'RegisterHook' now takes an additional param after the first param if the function has a return value. The param contains the return value pointer and you can use ':get()' to retrieve the value and ':set(<new-value>)' to set it but the return value of the actual Lua function will take precedence over calling 'set'.

Example:
```lua
RegisterHook("/Script/Engine.Character:CanJumpInternal", function()
    -- Some code.
end, function(ParamContext, ParamReturnValue)
    -- This value will be the original return value in the case where the pre-hook didn't modify it.
    -- It will be the modification done by the pre-hook otherwise.
    print(string.format("Return Value: %s\n", ParamReturnValue:get()))
end)
```


UE4SS:
This is ready to be merged.

PR #307, opened 2024-01-24
Added UEHelpers.FindOrAddFName

UE4SS:
This function will return the FName if it exists, and otherwise it will create it.

The UE documentation for 'EFindName::FNAME_Add' doesn't seem to be correct. It says: "Find a name or add it if it doesn't exist.", however, if the name already exists, it will add a new name with a number suffixed to it like "_0". This new UEHelpers function doesn't have this functionality and instead will return the original name without any extra suffix.

Also updated the docs to correctly point out that FNAME_Add is the default for the FName constructor, not FNAME_Find. Also added EFindName to the docs.

### DIFF
--- a/UE4SS/src/Mod/LuaMod.cpp
+++ b/UE4SS/src/Mod/LuaMod.cpp
@@ -3148,6 +3148,8 @@ Overloads:
         // Without this, the metatable won't be created until an FName is constructed by another part of UE4SS
         LuaType::FName::construct(lua, Unreal::FName(static_cast<int64_t>(0)));
         lua_setglobal(lua.get_lua_state(), "FName");
+        LuaType::FName::construct(lua, Unreal::NAME_None);
+        lua_setglobal(lua.get_lua_state(), "NAME_None");
         // FName Class -> END
 
         // FPackageName -> START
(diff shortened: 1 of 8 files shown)


UE4SS:
This fixes problems with BPModLoader and ConsoleEnabler where it couldn't find the mod class and it would set the console key to `F10_0` instead of `F10`.

UE4SS:
> I thought the _0 issue was caused by our sig for 5.1+ games finding the wrong fname ctor.
I think it's caused by either some change in the UE code or some undocumented behavior that for some reason didn't affect us before.
I believe this PR should solve it for all engine versions.

massclown:
I had the `FName` add `_0` behavior go away after manually specifying the correct constructor in signatures (it used to be a wrong overload of the `FName` constructor, I believe, both functions were pretty similar, same basic blocks essentially, same logic).

UE4SS:
I think either way, this PR is good because it will make these mods (and other mods using the same function) work regardless.

Buckminsterfullerene02:
> I think either way, this PR is good because it will make these mods (and other mods using the same function) work regardless.

So ready to merge?

UE4SS:
> > I think either way, this PR is good because it will make these mods (and other mods using the same function) work regardless.
> 
> So ready to merge?

In my opinion, yes.

PR #314, opened 2024-01-27
5.3 Support

localcc:


UE4SS:
@Buckminsterfullerene02 I don't have any knowledge surrounding the usmap generator so I can't tell you whether this PR is good or not.
I can tell you that I don't see any obvious generic mistakes but I can't tell you what effect this PR will have on the generated file.

Also this PR is probably poorly named since the only changes it makes are related to the usmap generator.

trumank:
> Also this PR is probably poorly named since the only changes it makes are related to the usmap generator.

Looks like it's part of the larger 503 support PR in the unreal submodule.

Buckminsterfullerene02:
Apologies that's my mistake. I requested a review from you here thinking that you'd look at the UE psuedo PR and comment about general 503 support of whatever here lol

PR #322, opened 2024-02-02
UI tweaks

localcc:
- feat(UE4SS): UI improvements


Buckminsterfullerene02:
Why is dump properties and objects not in the dumpers tab?

Buckminsterfullerene02:
Oh well if using the button is common that's fine, I totally understand. Though shouldn't there also be a keybind for that? 

My thought process is that it's not particularly common for most users (as far as I'm aware) while hot reloading mods is and it would be a bit annoying to accidentally misclick, but if dumping objects is fast (I can't remember and can't check) it wouldn't really matter.

Also, we will be releasing 3.0 today, so your changes will likely be for 3.1 instead. This PR is more of a quick last minute one to make the gui look a bit more interesting and visibly marks a major change between 2.x and 3.x.

UE4SS:
> Oh well if using the button is common that's fine, I totally understand. Though shouldn't there also be a keybind for that?
> 
> My thought process is that it's not particularly common for most users (as far as I'm aware) while hot reloading mods is and it would be a bit annoying to accidentally misclick, but if dumping objects is fast (I can't remember and can't check) it wouldn't really matter.
> 
> Also, we will be releasing 3.0 today, so your changes will likely be for 3.1 instead. This PR is more of a quick last minute one to make the gui look a bit more interesting and visibly marks a major change between 2.x and 3.x.

The time to dump objects is generally between less than a second and 5 seconds depending on the size of the game and the power of the users computer, maybe 10 seconds if the game is absolutely massive and the computer is absolutely terrible.

As far as convenience is concerned, I'm not against making the button somehow right-aligned so that it's not immediately next to the "Restart Mods" button.

Buckminsterfullerene02:
Just to clarify, I've not got a problem with the button being there. I think maybe aligning it to the right is fine, if not, the current layout is fine too. 

Sad to hear about the docking changes holdup, does seem like a nice change of/when it comes through.

PR #323, opened 2024-02-02
chore: bump version

localcc:


UE4SS:
I don't understand this PR.
I thought we had stopped using version.cache ?
I just checked and apparently we still use version.cache to populate the macros that we use to display the version in the log/console, but I could've sworn we stopped doing that, I must be remembering wrong.

It's also marked as UVTD but makes not changes in UVTD.

PR #325, opened 2024-02-02
fix(UE4SS): fix lua FString setter

localcc:
This fixes the FString setter from lua which added an additional `\0`
byte at the end with each assignment.

### DIFF
--- a/UE4SS/src/LuaType/LuaUObject.cpp
+++ b/UE4SS/src/LuaType/LuaUObject.cpp
@@ -1107,7 +1107,7 @@ namespace RC::LuaType
             {
                 auto lua_string = params.lua.get_string();
                 auto fstring = Unreal::FString{to_wstring(lua_string).c_str()};
-                string->SetCharArray(fstring.GetCharTArray());
+                *string = fstring;
             }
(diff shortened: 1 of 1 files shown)


UE4SS:
We should to ask @Okaetsu regarding this.
They were the one that added that extra null terminator to `SetCharArray`: https://github.com/Re-UE4SS/UEPseudo/commit/2a1ecc59

Okaetsu:
> We should to ask @Okaetsu about this. They were the one that added that extra null terminator to `SetCharArray`: [Re-UE4SS/UEPseudo@2a1ecc59](https://github.com/Re-UE4SS/UEPseudo/commit/2a1ecc59)

I have no issues with removing the null terminator from SetCharArray itself other than if you're manually creating a TCHAR Array and you forget a null terminator, things will break. I'm unsure if this would ever be a problem, but just something I thought of.

I'm thinking if it would make sense to add a check to see if a null terminator exists at the end and append one if necessary rather than always including it in SetCharArray.

UE4SS (approved):

Caminante3:
but, where be put the these solution code?

PR #327, opened 2024-02-02
Update readme and docs for 3.0

Buckminsterfullerene02:
need to update change log too but can't make two commits without discarding changes on GitHub mobile lol


assets/Changelog.md
```diff
@@ -120,10 +141,6 @@ Add additional extensions to USMap dumper - Atenfyr; Archengius
 
 Fix bug in USMap dumper with enums with 256 entries - Atenfyr
 
-### Live View
-
-### UHT Dumper
-
```
UE4SS, line 126: I wanted to keep empty sections so that it's clear exactly where to put things for the future.
Without them, people will keep re-creating and re-ordering them randomly leaving readers with no consistency for future changelogs.

Buckminsterfullerene02: Fine

PR #337, opened 2024-02-04
fix gui crash

localcc:
- chore: bump unreal submodule
- fix(UE4SS): fix GUI close crash


UE4SS (approved):
The code looks fine, though there's no explanation to why this would inconsistently cause a crash which is unsatisfying but if it fixes the problem then I think it's still good.

localcc:
from my testing the crash is always consistent when closing the gui, that's when imgui tries to free the font atlas it doesn't own and crashes

PR #372, opened 2024-02-09
Updated Lua doc for RegisterHook to accurately state how the callback params work

UE4SS:


Buckminsterfullerene02 (commented):
Shouldn't "native/non BP" be /Script/ and vice versa? Or at least mention that at the top. Remember that a lot of people following these dibs have no knowledge of ue so aren't nearly as likely to understand what these are than mapping them directly to how their string "looks"

UE4SS:
> Shouldn't "native/non BP" be /Script/ and vice versa? Or at least mention that at the top. Remember that a lot of people following these dibs have no knowledge of ue so aren't nearly as likely to understand what these are than mapping them directly to how their string "looks"

The problem is, I don't know if a function starting with `/Script/` is guaranteed to be native, and that not starting with `/Script/` guarantees that it's non-native.
If you can tell me that this is the case, then I'll change it.

Buckminsterfullerene02:
Native is always /Script/
Packages can either be /Engine/ for engine content or /Game/ for game content

Buckminsterfullerene02 (approved):

PR #387, opened 2024-02-13
Update changelog for 3.0.1

Buckminsterfullerene02:
**Description**

update version cache for 3.0.1

mini changelog FAQ because there's no where else to put it

**Type of change**

- Chore for new version release

**Additional context**

Anyone unhappy about there being a mini FAQ/PSA in the top of the changelog? Since the vast majority of traffic in the repo is directly to the releases page I feel like having this info at the top is by far the most visible for all users.


Buckminsterfullerene02:
Don't approve/merge yet I want to wait for #389 first

UE4SS:
I think it's fine to have a minified changelog at the top that has to most important and relevant information for most users.

Perhaps it would be a good idea to mention in each update if the ABI for C++ mods has been broken ?

Buckminsterfullerene02:
ABI? Do you mean API? Sure but has the C++ API been broken in this patch?

UE4SS:
> ABI? Do you mean API? Sure but has the C++ API been broken in this patch?

No I mean ABI: https://en.wikipedia.org/wiki/Application_binary_interface
For a C++ mod to work, the ABI of UE4SS has to match the version of UE4SS that the mod was built with, if it doesn't, the mod won't load at runtime.
Basically, if a patch (i.e. 3.0.1) doesn't have anything that breaks ABI then C++ mods that were built with the previous version (i.e 3.0.0) don't need to rebuild targeting the new version and release a new version of their mod for their users.
The versions used in the example above are just an example, and 3.0.1 does break ABI afaik.
Perhaps localcc could explain this in a neater way.

Buckminsterfullerene02:
Oh that's a good point, I think it is necessary to include that at the top too then. Yes ABI will have changed.

assets/Changelog.md
```diff
@@ -20,33 +40,54 @@ Added new docs pages for devlogs, C++ examples, C++ BP macros - Buckminsterfulle
...
 
 ### Live View
-Fix gui crash when clicking close window button ([#337](https://github.com/UE4SS-RE/RE-UE4SS/pull/337)) - localcc
+Fix GUI crash when clicking close window button ([UE4SS #337](https://github.com/UE4SS-RE/RE-UE4SS/pull/337)) - localcc
 
 ### UHT Dumper
 
 ### Lua API
-Fix FindObject and Mod:SetSharedVariable userdata type matching ([#342](https://github.com/UE4SS-RE/RE-UE4SS/pull/342)) - Lyrth
+Fix Lua FString setter ([UE4SS #325](https://github.com/UE4SS-RE/RE-UE4SS/pull/325)) - localcc
```
localcc, line 83: Yangff should actually be credited here

localcc (changes requested):

assets/Changelog.md
UE4SS: What's up with the requested change diff ?
It's saying to replace one fix with a completely different one, but your comment has nothing to do with that, leaving me a bit confused.

Buckminsterfullerene02: No you're right I'm really confused by this too

localcc (approved):

PR #412, opened 2024-02-27
Fix invalid BPLoader mod subdir behavior

ethangreen-dev:
This PR fixes a small error which prevented BPLoader from loading mods contained within subdirs of `LogicMods`.

Prior to this change, `LogicMods/Author-ModName/ModName.pak` would be _recognized_ by BPLoader but not actually loaded. If this file was moved into the top-level dir `LogicMods/ModName.pak`, it would then work.

### DIFF
--- a/assets/Mods/BPModLoaderMod/Scripts/main.lua
+++ b/assets/Mods/BPModLoaderMod/Scripts/main.lua
@@ -138,8 +138,15 @@ local function LoadModConfigs()
         end
     end
 
+    local Files = LogicModsDir.__files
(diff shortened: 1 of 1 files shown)


Buckminsterfullerene02:
Can you show an example where you have tested this please (I'd like to test myself but I cannot build UE4SS from source atm).

localcc:
Any status update on this?

Buckminsterfullerene02 (approved):
Finally got round to testing, this works fine.

ethangreen-dev:
Ah, sorry, got distracted with other things. Thanks for your patience and for merging this :-)

PR #425, opened 2024-03-05
fix: make FString a local lua param

### LINKED ISSUE #397: Use after free on values return by ufunctions in lua scripts (as they are assumed remote)
In `call_ufunction_from_lua`, `DynamicUnrealFunctionData dynamic_unreal_function_data` is an object on the stack used to hold the parameters of the called function as well as the return values. 
`m_property_value_pushers` is called to handle these return values.

The object is passed into the pusher as a data parameter, and subsequent processing is performed to obtain the lua object.
For example, in the FString  push_strproperty is called, the
```
string = static_cast<Unreal::FString*>(params.data);
```

Subsequently，
```
LuaType::FString::construct(params.lua, string);
```

Note that here `params.data` points to `dynamic_unreal_function_data.data[XXX]`, an address on the stack.
In `LuaType::FString`, this `Unreal::FString* unreal_object` is written directly to the `m_cpp_object` of the remoteobject. This causes it to point to an address on the stack. Despite the `lua.transfer_stack_object(...)` called here, it only acts on the `LuaType::FString`, and the remote object itself is still copied as a pointer.

This address is considered free after `call_ufunction_from_lua` returns, however the address is used in subsequent lua scripts that uses LuaType::FString and can be overwrite because it points to a stack address.

The observed is that `FString` returned values are volatile and cannot be accessed or get the wrong value after some operations.
My minimal reproduction code is
```
NotifyOnNewObject("/Script/Engine.PlayerState", function (ConstructedObject )
print(string.format("Constructed PlayerState: %s\n", ConstructedObject:GetFullName()))
local name = ConstructedObject:GetPlayerName())
print(string.format(" address :%d, name: %s\n", ConstructedObject:GetAddress(), name:ToString()))
--print(string.format(" address :%d, name: %s %s\n", ConstructedObject:GetAddress(), ConstructedObject:GetPlayerName():ToString(), name. ToString()))
end)
```

This will have the error output given after e5469a2. Although 892eefe...e5469a2 doesn't change the code much (or at all), so I think this problem also exists in the previous code, but maybe more stealthy.

Also, there is a memory leak problem here because the tarray inside the FString is never freed, as FString is always a pointer and UE4SS think UE will destruct it.

One possible solution is to use a LocalObject to store the FString, but I think more objects should be affected by it.. more than just the FString. All structures returned via the stack including the out parameter should be affected. 
And there should be a way to handle them while distinction between real remoteobject that manages by GC..

Also calling the correct destructor might have some troubles (except for classes like FString which you already have implement on ue4ss side)? in other caes, is FProperty::DestroyValue required?
image
image
image
image

Not using the template since I'm not even sure what is the minimum version has this problem..

localcc:
**Description**

This fixes some issues related to FString lifetime causing it to become
garbage data, FString should be stored by value instead.

Credits: @yangff
Please include a summary of the change and which issue is fixed. Include relevant motivation and context. List any dependencies that are required for this change.

Fixes #397

**Type of change**

**How Has This Been Tested?**

Tested on an Unreal 5.1 game with calling/hooking different functions returning FString, doing stuff on the stack and trying to read the value back.

**Checklist**

### DIFF
--- a/UE4SS/include/LuaType/LuaFString.hpp
+++ b/UE4SS/include/LuaType/LuaFString.hpp
@@ -16,7 +16,7 @@ namespace RC::LuaType
             return "FString";
         }
     };
-    class FString : public RemoteObjectBase<Unreal::FString, FStringName>
+    class FString : public LocalObjectBase<Unreal::FString, FStringName>
     {
       private:
         explicit FString(Unreal::FString* object);
--- a/UE4SS/src/LuaType/LuaFString.cpp
+++ b/UE4SS/src/LuaType/LuaFString.cpp
@@ -6,7 +6,7 @@
 
 namespace RC::LuaType
 {
-    FString::FString(Unreal::FString* object) : RemoteObjectBase<Unreal::FString, FStringName>(object)
+    FString::FString(Unreal::FString* object) : LocalObjectBase<Unreal::FString, FStringName>(std::move(Unreal::FString(*object)))
     {
     }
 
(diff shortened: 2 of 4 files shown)


UE4SS (approved):
The code changes look good to me.
I haven't tested it.

At some point I think we should rename `get_remote_cpp_object` & `get_local_cpp_object` to `get_cpp_object` because I don't think there's really any reason that we need do differentiate between them when retrieving the underlying object and the remote version could return a reference instead of a pointer to fix the syntax difference.

PR #429, opened 2024-03-08
fix(GUI): fix some widget alignments and font scaling issues

### LINKED ISSUE #420: [BUG - Release] - Increasing GUI scale hides toggle watch output button
**Branch or Release**
3.0.1 experimental build of #419 (but on release too)

**Game and Engine Version**
DRG

**Describe the bug**
I increased the GUI scale to 1.4 because the text is too small. Then in the watches tab, the toggle button that opens the watches text box disappears. Setting scale back to 1 brings it back again.

It seems like the more the scale, the further the button is hidden away. So at 1.2 scale the button is just a slither.

**Mods directory**
N/A

**To Reproduce**
1. Launch game
2. Change GUI scale to 1.4 or higher
3. Set watch
4. See that text box toggle button gone

**Expected behavior**
Text box toggle button still there at higher GUI scalings.

N/A

 - OS: Win10

**Additional context**
N/A

Lyrth:
### Description

Was supposed to only fix the watches tab, but I guess other parts of the GUI needed some tiny bits of attention too.

<details>
<summary><b>Fixes general alignment issues in the Console, Live View, and Watches tabs.</b></summary>

- Removed the weird gap underneath the Console log output, seems that it used to be meant for something else, but for now should probably fill that gap.
- Fixed the bottom panel of the LiveView splitter panels getting cut and extending beyond the window (seen with font scaling >= 1.0). I think one time I wondered if I'm missing a property entry because it might be hidden away outside the window. 
  - The bottom of the pane appears at around font scaling <= 0.7-ish though, so the spacing under the pane seem to vary depending on scaling. Fixed by using `ImGui::GetContentRegionAvail` instead of `GetContentRegionMaxAbs`, works with scaling.
- Made the bottom and right borders of the primary panes in Console, Live View, and Watches tab have the same distance from the window border. Tiny detail, but why not. Also works fine with scaling, *although* should this distance also scale with GUI scaling? Probably sounds too complicated (will also need to deal with other spacings, like between widgets).
</details>

<details>
<summary><b>Changes specific to the Live View tab</b></summary>

- Fixed the Copy Search Result button's alignment according to scaling. Used to go beyond the window for scaling > 1, and leaves some padding for scaling < 1. This requires a bit of text duplication though (not ImGui ID), so will need to change the string in two places if it needs to be edited (a var seems overkill? And putting a string somewhere else than where it's used in feels awkward, especially for immediate-mode layouting).
- Have the splitter panels scale when the window is resized. Now the panel height ratio will be kept; e.g. a 25% top and 75% bottom Live View layout will remain that way no matter the window height, unless you bring the window down to a very short size - each panel is limited to some minimum height (from `ImGui::GetFrameHeight`), so it'll just expand as 50%/50% when bringing the window back to normal size. Probably also fixes the times I get extremely thin panels, and the splitter going beyond the unreachable.
</details>

<details>
<summary><b>Changes specific to the Watches tab</b></summary>

- Put a space between the tab title's icon and name.
- Remove the table's `ImGuiTableFlags_NoPadOuterX` flag because it makes widgets touch the table borders.
- Controls column width fixed, also works with scaling.
- Put "Save" as the heading for the rightmost column (it was confusing what the checkbox is for at first), also width fixed.
- Replaced the watch history show/hide button with an `ImGui::Selectable`, as it's much simpler to use (it's a struggle to align text in a tiny button). Also used font icons for `+` and `-`. 
<details>
<summary><sub>my buton tex no senter :(</summary>

lyr_080324_KrB32tjsna
</details>

- History multiline text box is now `ImGui::GetTextLineHeight` times 10 lines high to account for font scaling, instead of a fixed 200px.
</details>

Fixes #420 

> [!NOTE]
> These changes were tested on Windows 10, 100% system DPI scaling, default title bar height of `31`. (Why the latter matters, I'll elaborate later at the bottom).
> Probably need to check if behavior changes on Windows 11 or other environments with a different window manager (Linux?), although I don't currently have access to either.

### Type of change

### How has this been tested?

Have GUI console enabled, and check the Console, Live View, and Watches tabs. Add two watches from any random properties. Tests done on font scale factors 2.0, 1.0, 0.4.

### Checklist

### Screenshots

<details>
<summary>Before/After</summary>

**0.4 scale**
0 4scale

**1.0 scale**
1scale

**2.0 scale**
2scale

</details>

<details>
<summary>Screen recording of Watches tab and Live View splitter resizing.

<b>Warning: some flashing.</b></summary>

https://github.com/UE4SS-RE/RE-UE4SS/assets/38674984/6ecf3195-3284-4464-8af8-d3eac01647da

</details>

### Notes

<details>
<summary><b><i>What do the numbers mean?</i></b></summary>

You'll often find `31` used throughout the sizes. This is Windows 10's default *native* title bar height (for me, at least).

For some reason ImGui does not reduce its own viewport height to compensate for the native title bar. So for example my entire window's height is 793px, border to border. The ImGui viewport height (inside the window, excluding the border and title bar) will still be 793px high, but the *actual* visible area is only `793px - 31px = 762px`, as if the entire ImGui viewport is shifted down and the bottom 31 pixels get hidden beyond the window border.

As for the numbers,
- `TextEditor`, `watch_render_frame`
  - `-16.0f` = `-8` left window margin + `-8` right window margin
  - `-31.0f + -8.0f` = `-31` title bar offset + `-8` bottom window margin
- `LiveView_InfoPanel`, `ImGui_Splitter`, `LiveView_TreeView`: `-16.0f` = `-8` left window margin + `-8` right window margin
- `##Search by name` via `ImGui::PushItemWidth`
  - `8.0f` = spacing between text box on the left and the button on the right
  - `16.0f` = `8` left window margin + `8` right window margin
  - `FramePadding.x * 2.0f` = button inside padding, left and right
  - `ImGui::CalcTextSize` = text width
- `split_pane_height`
  - `ImGui::GetContentRegionAvail().y` = remaining space from current pos to bottom of ImGui viewport (beyond the window)
  - `-31.0f + -8.0f` = `-31` title bar offset + `-8` bottom window margin
  - `-4.0f` = height of splitter bar
- `Controls` column
  - `ImGui::GetFrameHeight * 3.0f` = three square controls
  - `4.0f` = 2.0f px spacers between controls, x2
- `Save` column: `ImGui::GetFrameHeight` = one square control
- Each `##history` textbox
  - `-2.0f` = right margin from column right edge
  - `ImGui::GetTextLineHeight() * 10.0f` 10 text lines + `ImGui::GetStyle().FramePadding.y * 2.0f` top and bottom padding

</details>

> [!NOTE]
> Currently, the pull requests makes an assumption that the title bar-caused offset will always be 31. On the unfortunate case that the title bar is seen as less than 31px high, or even zero (on Linux maybe), the most severe effect should *only* be padding appearing on the bottom. This is only my assumption though, I likely don't know of edge cases where ImGui behaves in a weirder way regarding this in more specific environments.

<details>
<summary><sub><sup>p.s.</sup></sub></summary>

<sub>Pardon the overly detailed post for this small PR. Was bored and felt like doing a semi-technical document lol</sub>
</details>

### DIFF
--- a/UE4SS/src/GUI/Console.cpp
+++ b/UE4SS/src/GUI/Console.cpp
@@ -128,11 +128,10 @@ namespace RC::GUI
         //*/
 
         /**/
-        const float footer_height_to_reserve = (ImGui::GetStyle().ItemSpacing.y * 10.0f) + ImGui::GetFrameHeightWithSpacing();
-        {
-            std::lock_guard<std::mutex> guard(m_lines_mutex);
-            m_text_editor.Render("TextEditor", {-10, -footer_height_to_reserve});
-        }
+
+        std::lock_guard<std::mutex> guard(m_lines_mutex);
+        m_text_editor.Render("TextEditor", {-16.0f, -31.0f + -8.0f});
+
         ImGui_AutoScroll("TextEditor", &m_previous_max_scroll_y);
         //*/
     }
--- a/UE4SS/src/GUI/GUI.cpp
+++ b/UE4SS/src/GUI/GUI.cpp
@@ -144,7 +144,7 @@ namespace RC::GUI
                     should_unset_listeners = true;
                 }
 
-                if (ImGui::BeginTabItem(ICON_FA_EYE "Watches"))
+                if (ImGui::BeginTabItem(ICON_FA_EYE " Watches"))
                 {
                     listeners_are_required = true;
                     m_live_view.render_watches();
--- a/UE4SS/src/GUI/LiveView.cpp
+++ b/UE4SS/src/GUI/LiveView.cpp
@@ -2584,7 +2584,7 @@ namespace RC::GUI
 
     auto LiveView::render_info_panel() -> void
     {
-        ImGui::BeginChild("LiveView_InfoPanel", {-14.0f, m_bottom_size}, true, ImGuiWindowFlags_HorizontalScrollbar);
+        ImGui::BeginChild("LiveView_InfoPanel", {-16.0f, m_bottom_size}, true, ImGuiWindowFlags_HorizontalScrollbar);
 
         size_t next_object_index_to_select{};
 
@@ -2819,7 +2819,9 @@ namespace RC::GUI
         {
             ImGui::BeginDisabled();
         }
-        ImGui::PushItemWidth(-160.0f);
+
+        // Update this text if corresponding button's text changes. Textinput width = Spacing + Window margin + Button padding + Button text width
+        ImGui::PushItemWidth(-(8.0f + 16.0f + ImGui::GetStyle().FramePadding.x * 2.0f + ImGui::CalcTextSize(ICON_FA_COPY " Copy search result").x));
         bool push_inactive_text_color = !m_search_field_cleared;
         if (push_inactive_text_color)
         {
@@ -3088,6 +3090,8 @@ namespace RC::GUI
         }
 
         ImGui::SameLine();
+
+        // Remember to update text width calculations for the last ImGui::PushItemWidth call if this text gets updated.
         if (ImGui::Button(ICON_FA_COPY " Copy search result"))
         {
             StringType result{};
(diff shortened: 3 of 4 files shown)


aoqia194:
I really hope this addresses the issue with watches and how most of the watches table is cut off. If it does then this is so poggers and if not I can include a screenshot example.

Buckminsterfullerene02:
Need to test myself but this is indeed poggers, such as the properties view in live view section being cut off at the bottom, that was already really annoying.

UE4SS:
This looks good with Proton on Linux.
I don't feel like merging this with the Linux PR to see if it looks good natively.

Buckminsterfullerene02 (approved):
Works great!

PR #436, opened 2024-03-13
Fix off-by-one error in TArray::prepare_to_handle 

### LINKED ISSUE #435: [BUG - RELEASE] Lua - accessing new indexes in TArray does not correctly resize/zero memory
**Branch or Release**
v3.0.1

**Game and Engine Version**
Multiple, but confirmed local repro in Pal World v0.5.1 / UE 5.1.1

**Describe the bug**
Accessing new indexes in `TArray` only zeroes / resizes to `new index - 1`, causing trailing zeroed values when appending to a `TArray`

**Mods directory**
A zip of your Mods folder if you've installed any mods that don't come with UE4SS by default

**To Reproduce**
See comment reference below.

**Expected behavior**
See comment reference below.

If applicable, add screenshots to help explain your problem, and upload the log and .dmp files located in the Win64 directory.

 - OS: Win11

**Additional context**
Forked from #378 as it's a distinct issue from the original report

              Tested TArray, and unfortunately the same hack doesn't apply, as it seems that the parameter passed into the hook is a copy of the original ('default' TArray?), and therefore the changes don't get reflected. A TArray being passed *as a copy* in a hook could possibly be a separate issue though, although I'm not sure if it work the same way if the UFunction was called natively (with a proper TArray argument) instead of from Lua.

```lua
local kal = StaticFindObject("/Script/Engine.Default__KismetArrayLibrary") ---@cast kal UKismetArrayLibrary


local handler = function() end

--- int32 Array_Length(const TArray<int32>& TargetArray)
local pre0, post0 = RegisterHook("/Script/Engine.KismetArrayLibrary:Array_Length", function(...) handler(...) end)

local arr = { 1, 2, 3, 4 }

handler = function(p0, p1)
    local tarray = p1:get()
    print(#tarray)              -- prints 0
    print(tarray:GetArrayNum()) -- prints 0
    print(tarray:GetArrayMax()) -- prints 0

    for i = 1, #arr do  -- loops 4 times
        tarray[i] = arr[i]
        print(i..": "..tarray[i]) -- prints "1: 1"..."4: 4"
    end

    print(#tarray)              -- prints 3 (wrong, off-by-one?)
    print(tarray:GetArrayNum()) -- prints 3 (wrong, off-by-one?)
    print(tarray:GetArrayMax()) -- prints 6
end

-- prints: 0  (wrong)
print(tostring(kal:Array_Length({})))

UnregisterHook("/Script/Engine.KismetArrayLibrary:Array_Length", pre0, post0)


------------------------------------


local handler = function() end

--- int32 Array_IsValidIndex(const TArray<int32>& TargetArray, int32 IndexToTest)
local pre1, post1 = RegisterHook("/Script/Engine.KismetArrayLibrary:Array_IsValidIndex", function(...) handler(...) end)

local arr = { 1, 2, 3, 4 }

handler = function(p0, p1, p2)
    local tarray = p1:get()
    print("IndexToTest: "..p2:get()) -- prints "IndexToTest: 1"

    for i = 1, #arr do  -- loops 4 times
        tarray[i] = arr[i]
        print(i..": "..tarray[i]) -- prints "1: 1"..."4: 4"
    end
end

-- prints: false  (wrong)
print(tostring(kal:Array_IsValidIndex({}, 1)))

UnregisterHook("/Script/Engine.KismetArrayLibrary:Array_IsValidIndex", pre1, post1)
```

_Originally posted by @Lyrth in https://github.com/UE4SS-RE/RE-UE4SS/issues/378#issuecomment-1937753705_

dnchattan:
**Description**
Fixes #435 (forked from it's original issue). The new array size is incorrectly computed in `TArray::prepare_to_handle` and targets a size that is 1 less than the intended array size. This causes it to not zero out and resize except for when you add _additional_ elements after it- this leaves a trail of zeroed out entries, with one "missing" because the size was off for each call.

**Type of change**

**How Has This Been Tested?**

Verified the following script reproduces the issue and outputs garbage for [4] and [5] the first time, and 0x0000 for [4] the second time ([5] still garbage).  Updated code and tested script to confirm both iterations output 0x0000 for [4] and [5]. Additionally stepped through each iteration in the debugger to confirm the correct calls and values were being made.

```lua
print("TEST SCRIPT LOADED")

function printBaseId(ar, idx)
    local campId = ar[idx]
    print(string.format("BaseCampIds[%d] 0x%X%X%X%X", idx, campId.A, campId.B, campId.C, campId.D))
end

function printBases()
    PalGroupGuildBase = FindFirstOf("PalGroupGuildBase")
    if PalGroupGuildBase == nil then
        print("PalGroupGuildBase not found")
        return
    end
    print("-- dump BaseCampIds[1-5] --")
    for idx = 1, 5 do
        printBaseId(PalGroupGuildBase.BaseCampIds, idx)
    end
    print("-- repeat --")
    print("-- dump BaseCampIds[1-5] --")
    for idx = 1, 5 do
        printBaseId(PalGroupGuildBase.BaseCampIds, idx)
    end
    
end

RegisterKeyBind(Key.Z, printBases)
```

**Checklist**

**Additional context**

### DIFF
--- a/UE4SS/src/LuaType/LuaTArray.cpp
+++ b/UE4SS/src/LuaType/LuaTArray.cpp
@@ -219,7 +219,12 @@ namespace RC::LuaType
     auto TArray::prepare_to_handle(const LuaMadeSimple::Type::Operation operation, const LuaMadeSimple::Lua& lua) -> void
     {
         auto& lua_object = lua.get_userdata<TArray>();
-        int64_t array_index = lua.get_integer() - 1; // Subtracting 1 here to account for that fact that Lua tables are 1-indexed
+        int64_t array_index64 = lua.get_integer() - 1; // Subtracting 1 here to account for that fact that Lua tables are 1-indexed
+        if (array_index64 < 0 || array_index64 > std::numeric_limits<int32_t>::max())
+        {
+            lua.throw_error("TArray index out of range.");
+        }
+        int32_t array_index = static_cast<int32_t>(array_index64);
 
         auto array = lua_object.get_remote_cpp_object();
         auto num = array->Num();
--- a/UE4SS/src/LuaType/LuaTArray.cpp
+++ b/UE4SS/src/LuaType/LuaTArray.cpp
@@ -223,10 +223,10 @@ namespace RC::LuaType
 
         auto array = lua_object.get_remote_cpp_object();
         auto num = array->Num();
-        if (array_index > num || num == 0)
+        if (array_index >= num || num == 0)
         {
             // We've verified with the above if-statement that we're not out-of-bounds for an int32.
-            int32_t count = array_index == 0 || num == 0 ? 1 : static_cast<int32_t>(array_index - num);
+            int32_t count = array_index == 0 || num == 0 ? 1 : static_cast<int32_t>(array_index - num + 1);
             lua_object.get_remote_cpp_object()->AddZeroed(count, lua_object.m_inner_property->GetElementSize(), Unreal::DEFAULT_ALIGNMENT);
         }
 
--- a/assets/Changelog.md
+++ b/assets/Changelog.md
@@ -1,5 +1,6 @@
 v3.1.0
 ==============
+
 TBD
 
 ## New
@@ -16,7 +17,6 @@ TBD
 
 ### Experimental
 
-
 ## Changes
 
 ### General
@@ -26,16 +26,19 @@ TBD
 ### UHT Dumper
 
 ### Lua API
+
 `print` now behaves like vanilla Lua (can now accept zero, one, or multiple arguments of any type).
 
 ### C++ API
 
 ### Repo & Build Process
 
-
 ## Fixes
 
 ### General
+
+Fixed adding elements to TArray in Lua incorrectly resizing and zeroing out previous values
+
 Fixed BPModLoaderMod giving "bad conversion" errors.
 
 Fixed some debug GUI layout alignments, especially with different GUI font scaling settings.
@@ -45,6 +48,7 @@ Fixed some debug GUI layout alignments, especially with different GUI font scali
 ### UHT Dumper
 
 ### Lua API
+
 Fixed FString use after free.
 
 Fixed the "IterateGameDirectories" global function throwing "bad conversion" errors.
@@ -53,10 +57,10 @@ Fixed `FText` not working as a parameter (in `RegisterHook`, etc.).
 
 ### C++ API
 
-
 ## Settings
 
 ### Added
+
 ```
 [Hooks]
 HookLoadMap = 1
@@ -64,9 +68,9 @@ HookLoadMap = 1
 
 ### Removed
 
-
 v3.0.1
 ==============
+
 2024-02-14
 
 This is a patch release for 3.0.0. Installing over 3.0 is as simple as replacing all files except for your `UE4SS-settings.ini` if you have custom configs.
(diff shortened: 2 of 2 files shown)


UE4SS/src/LuaType/LuaTArray.cpp
```diff
@@ -223,10 +223,10 @@ namespace RC::LuaType
 
         auto array = lua_object.get_remote_cpp_object();
         auto num = array->Num();
-        if (array_index > num || num == 0)
+        if (array_index >= num || num == 0)
         {
             // We've verified with the above if-statement that we're not out-of-bounds for an int32.
```
bitonality, line 228: Is this comment misleading? `array_index` is `int64_t` and `array-Num()` returns `int32` so technically the result of `array_index - num` would be `long long`? Overflow?

Lyrth: Regarding the current indexing, although not that relevant to the current PR, would it be better to
- Differentiate between a ::Get vs ::Set operation
- If it's a Get, disallow indexing outside [0, Num) and just return Lua `nil` (or nullptr UObject to be consistent with other UE4SS functions) when accessing outside the array
- If it's a Set, only allow [0, Num), plus the index after the last element for extending the array, to be set with a value, to avoid holes (if holes are required then the user must explicitly set each entry upwards, probably with `nil` or an empty). For indexes beyond that range, throw an error. Of course, this would be breaking.

UE4SS: This seems reasonable.

assets/Changelog.md
UE4SS, line 1: Some automatic formatting appears to have been applied to this file.
Please undo it.

UE4SS/src/LuaType/LuaTArray.cpp
bitonality:
```lua
local tarray = p1:get()
tarray[9223372036854775807] -- max signed int64
```
```cpp
        if (array_index >= num || num == 0)
        {
            // We've verified with the above if-statement that we're not out-of-bounds for an int32.
            int32_t count = array_index == 0 || num == 0 ? 1 : static_cast<int32_t>(array_index - num + 1);
            // count = -5...
            lua_object.get_remote_cpp_object()->AddZeroed(count, lua_object.m_inner_property->GetElementSize(), Unreal::DEFAULT_ALIGNMENT);
        }
```

Additionally, when something like `tarray[9999999999999999999]` is called then cpp `int64_t test = lua.get_integer()` returns `-1`. 

My previous concerns about overflow were in regards to a ::Get() with an index greater than maxint32...

Thoughts?

UE4SS: I think the bounds check I proposed above should fix this, but perhaps it should be promoted to an exception instead of a default of 0.

bitonality: Ah, didn't see the code included with your comment! Bounds check and explicit `int32_t` seem reasonable.

dnchattan: I agree this should be an exception rather than default of 0. Will adjust the code accordingly.

just curious, have you considered using safeint (or something similar) for bounds checking?

dnchattan: @UE4SS is there value in submitting this as a patch fix and follow with a separate breaking change for the new `:get()`/`:set()` limitations proposed-- or should I just do that as part of this PR?

I'm assuming all together makes sense unless you do something special to roll patches out faster and batch breaking changes together.

UE4SS: > just curious, have you considered using safeint (or something similar) for bounds checking?

I don't know what safeint is, I've never heard of it before.



> @UE4SS is there value in submitting this as a patch fix and follow with a separate breaking change for the new `:get()`/`:set()` limitations proposed-- or should I just do that as part of this PR?
> 
> I'm assuming all together makes sense unless you do something special to roll patches out faster and batch breaking changes together.

I'm not part of the decision making group so this is just my opinion on how I think it should be done.
Please ask on the discord server if you want an official answer.

As far as I'm aware, there's no procedure in place for taking advantage of separating a breaking change and a non-breaking change into two PRs, but the people over at the discord server would know more about that since I believe internal planning and procedures are done there.

We like to keep PRs as clean as possible but at the same time we need to be reasonable with how picky we are.
I think we should accept expanding PR scopes when it makes sense as long as the PR is re-named and re-described appropriately or possibly re-created if the changes fully supersede the original PR.
If not re-created a comment should be made so that people are notified to re-read and re-review because I don't think edits send notifications.
New changes should be added to the change log when appropriate as well, something that might easily be forgotten when an existing PR is expanded instead of a new one being created.

Buckminsterfullerene02: I think breaking changes should be kept seperate when it's not really that relevant to the existing PR. Usually we want to keep smaller breaking changes like this seperate to fixes so that fixes can go into patch release bumps and small breaking changes can go into minor release bumps. 

Since the next release will include build system change, UEPlatform and potentially Linux port, that will certainly be a minor release bump at least. While we could make an exception right here to say, "oh well it's not going to be a patch bump next release let's just bundle it in", I don't think we should set the tone on frequent exceptions just because lazy. It's not like PRs are difficult to make. I hope this makes sense! But I have indeed brought up the topic for discussion in the internal dev channel.

dnchattan: @Buckminsterfullerene02 I agree with that methodology, and is my preference as well to have more smaller scoped changes; it also means you can decide when you want to merge that change independently from this bugfix :)

UE4SS (approved):

PR #442, opened 2024-03-25
Made it easier for custom AActor classes in C++ mods to get notified on tick

UE4SS:
**Description**

This PR requires #443 to be merged, and https://github.com/Re-UE4SS/UEPseudo/pull/83 is also recommended.

This was done with the addition of the 'ENABLE_CLASS_RUNTIME_FUNCTIONALITY' macro, to be used in 'on_unreal_init' by C++ mods. The macro takes a C++ class as a param.
We use a general AActor::Tick hook to iterate all AActor classes that want a tick notification.

The macro is meant to be generic to allow more features to be added to it in the future.

A discussion needs to be had internally and also ideally with C++ mod creators regarding whether the `Tick` function should be called before or after the real `Tick` function that UE calls.
If it's called before, then any changes that the C++ mod applies in its Tick function will be able to affect the original Tick function.
If it's called after, then any changes that the C++ mod applies will override whatever the original Tick function did.
It's currently being called after the original Tick function.

Note that if the C++ class passed to the macro doesn't have a `Tick` function, the macro is still valid and will do nothing.
This is by design so that we can have more features tied to this macro in the future.

This PR will remain as a draft until #443 is merged and the question about when the Tick function should be called has been resolved.

Usage example:
```c++
// UE4SS_Main51Character.hpp
class AUE4SS_Main51Character : public Unreal::AActor
{
    DECLARE_EXTERNAL_OBJECT_CLASS(AUE4SS_Main51Character, UE4SS_Main51)

public:
    // Will automatically be called on every actor tick if we use the macro in 'on_unreal_init'.
    void Tick(float DeltaSeconds)
    {
        Output::send(STR("UE4SS_Main51Character Tick\n"));
    }
};

// UE4SS_Main51Character.cpp
IMPLEMENT_EXTERNAL_OBJECT_CLASS(AUE4SS_Main51Character)

// main.cpp
class MyAwesomeMod : public RC::CppUserModBase
{
    MyAwesomeMod() : CppUserModBase()
    {
        ModName = STR("MyAwesomeMod");
        ModVersion = STR("1.0");
        ModDescription = STR("This is my awesome mod");
        ModAuthors = STR("UE4SS Team");
    }

    void on_unreal_init() override
    {
        ENABLE_CLASS_RUNTIME_FUNCTIONALITY(AUE4SS_Main51Character)
    }
};
```

**Type of change**

**How Has This Been Tested?**

I created a C++ mod with a custom ACharacter class with a Tick function (similar to the example above), and verified that the Tick function was being executed by outputting something to the console.

**Checklist**


UE4SS:
Updated description to add that #443 needs to be merged before or alongside this PR.

UE4SS:
Note that the `Tick` function will only be called if the game devs made the Actor actually tick.
For BPs, doing something in `Event Tick` in the `Event Graph` is enough to make an actor tick, assuming that the actor has ticks enabled in it's `Class Defaults` section or elsewhere.

UE4SS:
All of the prerequisite PRs have been merged.
Now we just need to decide whether `Tick `should be called before or after the original Tick function.

I personally thing our `Tick` function should be called after the original one to give final authority over actor changes (i.e. changing properties) to the mod.

narknon:
I am hesitant to even have this as an option as opposed to using a tick function on a Singleton. This just seems like a performance nightmare if even a single mod hooks it, even if they only execute their logic on a specific actor.

UE4SS:
> I am hesitant to even have this as an option as opposed to using a tick function on a Singleton. This just seems like a performance nightmare if even a single mod hooks it, even if they only execute their logic on a specific actor.

What do you mean by "using a tick function on a Singleton" ?
But yes, it could be performance heavy, some profiling should probably be done.
Not sure how performance heavy our internal hook code actually is.

narknon:
Hooking an object with a tick function that is always used, but of which only one object exists. This should probably be part of a larger overhaul to sync UE4SS and the engine. I'm just not sure if the software is ready from a performance standpoint for a detour of actor tick, particularly if ever used from Lua.

UE4SS:
> Hooking an object with a tick function that is always used, but of which only one object exists. This should probably be part of a larger overhaul to sync UE4SS and the engine. I'm just not sure if the software is ready from a performance standpoint for a detour of actor tick, particularly if ever used from Lua.

How would you go about singling out a specific tick function ?
As far as I'm aware, they aren't exposed as a UFunction so we can't find them that way.
Also, this isn't exposed to Lua atm, and we should probably keep it that way for the moment until we have proof that it won't affect performance noticeably from just existing.

narknon:
Viewport or GameEngine are the most likely candidates. GameEngine probably has a lower chance of the vtable being changed by engine customizations.

UE4SS:
> Viewport or GameEngine are the most likely candidates. GameEngine probably has a lower chance of the vtable being changed by engine customizations.

Those aren't actor ticks though, so won't carry the implications expected of an actor tick, nor would you have cheap & easy access to your actor in there.
It sounds to me like it's a completely different feature.

nexustheru:
there are usally three ticks available,, before, ingame ticks and after. overloading the the main tick is never good. better do it preload or after load.

UE4SS:
> there are usally three ticks available,, before, ingame ticks and after.

Link to this ?
I know ticks happen at different times, but that's configured per actor and is outside the scope of this PR.
I just want C++ mods to be able to do things when the actor ticks.
Whether it's a PrePhysics tick or whatever isn't the important part here, that's a game dev decision.

> overloading the the main tick is never good. better do it preload or after load.

I don't know what you mean by "overload".
Perhaps you've misunderstood this PR ?
This PR isn't "overloading" any function, it's firing a C++ mod function immediately after the original actor tick has happened.

nexustheru:
> > there are usally three ticks available,, before, ingame ticks and after.
> 
> Link to this ? I know ticks happen at different times, but that's configured per actor and is outside the scope of this PR. I just want C++ mods to be able to do things when the actor ticks. Whether it's a PrePhysics tick or whatever isn't the important part here, that's a game dev decision.
> 
> > overloading the the main tick is never good. better do it preload or after load.
> 
> I don't know what you mean by "overload". Perhaps you've misunderstood this PR ? This PR isn't "overloading" any function, it's firing a C++ mod function immediately after the original actor tick has happened.

heres one https://docs.unrealengine.com/4.26/en-US/API/Runtime/PreLoadScreen/IPreLoadScreen/Tick/

nexustheru:
https://docs.unrealengine.com/4.26/en-US/ProgrammingAndScripting/ProgrammingWithCPP/UnrealArchitecture/Actors/Ticking/

UE4SS:
> https://docs.unrealengine.com/4.26/en-US/ProgrammingAndScripting/ProgrammingWithCPP/UnrealArchitecture/Actors/Ticking/

Yeah it's what I thought it was, and it's not relevant to this PR.

nexustheru:
https://www.casualdistractiongames.com/post/2018/12/10/unreal-tick-functions-delta-time-and-the-task-graph

nexustheru:
game itself are a seperate process logical, theres world ticks, game ticks whatever its called in diffrent game, and thiese instances loads upp game behavior. why not use this to add plugin behavior. it comes part of loading

nexustheru:
this way you can load game up even if plugin is crashing

UE4SS:
I don't understand what you're trying to say.
I don't think you understand this PR.
This PR isn't trying to add any kind of plugin behavior, it's just trying to give C++ mods an easy way to hook into an Actors tick function.

nexustheru:
> I don't understand what you're trying to say. I don't think you understand this PR. This PR isn't trying to add any kind of plugin behavior, it's just trying to give C++ mods an easy way to hook into an Actors tick function.

i do understand, and to be able to do this u need btw get around the protection, and when u done that this adress changes adress probarly. so keep track on the actor pawn controlled by the player and read its methods by tick, bc obuisly this changes to, no supported action, to remove this protection. denuvo. delving further would indeed impose the right. are you sure you want to do this? better leave this unsolved

narknon:
This has nothing to do with denuvo or any kinds of protections. UE4SS does not support bypassing any protections or modding of online competitive games.

UE4SS:
> > I don't understand what you're trying to say. I don't think you understand this PR. This PR isn't trying to add any kind of plugin behavior, it's just trying to give C++ mods an easy way to hook into an Actors tick function.
> 
> i do understand, and to be able to do this u need btw get around the protection, and when u done that this adress changes adress probarly. so keep track on the actor pawn controlled by the player and read its methods by tick, bc obuisly this changes to, no supported action, to remove this protection. denuvo. delving further would indeed impose the right. are you sure you want to do this? better leave this unsolved

You're not making any sense.
This PR is in a working condition, and denuvo shouldn't have any effect on it, nor do I care even if it does because bypassing DRM isn't what we're about here.

nexustheru:
"nor do I care even if it does because bypassing DRM isn't what we're about here."  ok , my mistake do go on.

nexustheru:
> This has nothing to do with denuvo or any kinds of protections. UE4SS does not support bypassing any protections or modding of online competitive games.

well isnt it what it does,, it simply aply to engine of unreal to inject out code. and i know games do have denuvos, hences why this dont always work bc pointer get all scrambled. i used IDA . walk around this issues by altering the needed dll instead of exe itself.

nexustheru:
> > This has nothing to do with denuvo or any kinds of protections. UE4SS does not support bypassing any protections or modding of online competitive games.
> 
> well isnt it what it does,, it simply aply to engine of unreal to inject out code. and i know games do have denuvos, hences why this dont always work bc pointer get all scrambled. i used IDA . walk around this issues by altering the needed dll instead of exe itself.

and i know u already dont know what im saying but someone does. last 8 post i sent u dont know anything what im saying.

narknon:
> > This has nothing to do with denuvo or any kinds of protections. UE4SS does not support bypassing any protections or modding of online competitive games.
> 
> well isnt it what it does,, it simply aply to engine of unreal to inject out code. and i know games do have denuvos, hences why this dont always work bc pointer get all scrambled. i used IDA . walk around this issues by altering the needed dll instead of exe itself.


No, this PR is not related to denuvo or injecting out code in the slightest. Please stay on topic for the PR.

UE4SS:
Closing because of inactivity.

PR #446, opened 2024-03-28
Fixed keybinds created from C++ mods executing multiple times

UE4SS:
**Description**

This only happened if the hot-reload feature was used to reload all mods.

This was fixed by introducing `CppUserModBase::register_keydown_event`.
It's identical to the one in UE4SSProgram, except it keeps track of which C++ mod registered the event so that it can be "unregistered" when the mod is reloaded.

Mods that continue to use the UE4SSProgram version will continue to have the same problem.

**Type of change**

**How Has This Been Tested?**

I created a C++ mod that registered a keybind with the new function that output something to the console, and verified that only one message in the console appeared after using the bind after spamming the "Restart All Mods" button.

**Checklist**

### DIFF
--- a/UE4SS/include/Mod/CppUserModBase.hpp
+++ b/UE4SS/include/Mod/CppUserModBase.hpp
@@ -6,6 +6,7 @@
 #include <Common.hpp>
 #include <File/Macros.hpp>
 #include <GUI/GUITab.hpp>
+#include <Input/Handler.hpp>
 
 namespace RC
 {
@@ -126,5 +127,8 @@ namespace RC
 
       protected:
         RC_UE4SS_API auto register_tab(std::wstring_view tab_name, GUI::GUITab::RenderFunctionType) -> void;
+        RC_UE4SS_API auto register_keydown_event(Input::Key, const Input::EventCallbackCallable&, uint8_t custom_data = 0) -> void;
+        RC_UE4SS_API auto register_keydown_event(Input::Key, const Input::Handler::ModifierKeyArray&, const Input::EventCallbackCallable&, uint8_t custom_data = 0)
+                -> void;
     };
 } // namespace RC
--- a/UE4SS/include/UE4SSProgram.hpp
+++ b/UE4SS/include/UE4SSProgram.hpp
@@ -62,8 +62,21 @@ namespace RC
         uint64_t safety_padding[8]{0};
     };
 
+    struct KeyDownEventData
+    {
+        // Custom data from the C++ mod.
+        // The 'custom_data' variable to UE4SSProgram::register_keydown_event will be used to determine the type of custom_data2.
+        uint8_t custom_data{};
+
+        // The C++ mod that created this event.
+        CppUserModBase* mod{};
+    };
+
     class UE4SSProgram : public MProgram
     {
+      public:
+        friend class CppUserModBase; // m_input_handler
+
       public:
         constexpr static wchar_t m_settings_file_name[] = L"UE4SS-settings.ini";
         constexpr static wchar_t m_log_file_name[] = L"UE4SS.log";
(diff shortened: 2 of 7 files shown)


Buckminsterfullerene02:
Can you send the code responsible for registering the keybind? I'm trying to replicate but am having troubles with it due to getting "failed to load dll" due to that part of the code being erreonous somehow (removing that code causes the mod.dll to load fine). Admittedly I have never written ue4ss mod code before as xmake branch is the first time I can actually compile ue4ss since c++ api was added.

I figured the following code would work inside of the mod's constructor:
```cpp
        const Input::EventCallbackCallable callback = [&] {
            Output::send<LogLevel::Verbose>(STR("P key was pressed\n"));
        };

        register_keydown_event(Input::P, callback);
```

UE4SS:
> Can you send the code responsible for registering the keybind? I'm trying to replicate but am having troubles with it due to getting "failed to load dll" due to that part of the code being erreonous somehow (removing that code causes the mod.dll to load fine). Admittedly I have never written ue4ss c++ mod code before as xmake branch is the first time I can actually compile ue4ss since c++ api was added.
> 
> I figured the following code would work inside of the mod's constructor:
> 
> ```c++
>         const Input::EventCallbackCallable callback = [&] {
>             Output::send<LogLevel::Verbose>(STR("P key was pressed\n"));
>         };
> 
>         register_keydown_event(Input::P, callback, 2); // I assume custom_data is 2 because it means the bind came from c++
> ```

This code is functioning, but wrong in the assumption that the last param should be 2.
The last param isn't needed and actually means nothing in this context.
The "2" that designates the bind as coming from a C++ mod is done automatically in the CppUserModBase wrapper for register_keydown_event that you've called in your code.

As far as your problem is concerned: Are you sure that you moved both UE4SS.dll and your mod dll after building ?
I tried your code and it works fine for me.

Buckminsterfullerene02:
> > Can you send the code responsible for registering the keybind? I'm trying to replicate but am having troubles with it due to getting "failed to load dll" due to that part of the code being erreonous somehow (removing that code causes the mod.dll to load fine). Admittedly I have never written ue4ss c++ mod code before as xmake branch is the first time I can actually compile ue4ss since c++ api was added.
> > I figured the following code would work inside of the mod's constructor:
> > ```c++
> >         const Input::EventCallbackCallable callback = [&] {
> >             Output::send<LogLevel::Verbose>(STR("P key was pressed\n"));
> >         };
> > 
> >         register_keydown_event(Input::P, callback, 2); // I assume custom_data is 2 because it means the bind came from c++
> > ```
> 
> This code is functioning, but wrong in the assumption that the last param should be 2. The last param isn't needed and actually means nothing in this context. The "2" that designates the bind as coming from a C++ mod is done automatically in the CppUserModBase wrapper for register_keydown_event that you've called in your code.
> 
> As far as your problem is concerned: Are you sure that you moved both UE4SS.dll and your mod dll after building ? I tried your code and it works fine for me.

Ahhh yeah I was forgetting to replace the UE4SS.dll as well! Apologies! Thanks for the explanation though, I appreciate it. I can replicate your fix, so I think this is fine.

btw, I searched at least 30 ue4ss cpp mods for this function (which already existed under the same name but only through `UE4SSProgram::`) using [this search link](https://github.com/search?q=language%3Acpp+CppUserModBase+NOT+repo%3AUE4SS-RE%2FRE-UE4SS+NOT+repo%3AEpicGames%2FUnrealEngine&type=code&p=2) but I didn't find a single person using this function. I'd imagine people would actually start using it now that it's in the mod base but I'm sure there are probably a bunch more things not being used, which goes to show how much we need proper c++ api docs generation of some sort. Maybe I'll look into it when I'm less busy irl.

Buckminsterfullerene02 (approved):

Issue #370, opened 2024-02-08
[BUG - Release] 3.0 Pre and Post begin play logic duplicated for every mods(checked with AddComponent node)


MMadmer:
**Branch or Release**
3.0

**Game and Engine Version**
5.1.1

**Describe the bug**
If u have logic into ur PostBeginPlay or PreBeginPlay then the logic will be worked out for every ModActor mods.

**To Reproduce**
1. Put more than 1 BP mod.
2. Create PostBeginPlay and use AddComponent to any not ModActor actor.
3. Get components number from target actor.

UE4SS:
I don't understand the problem, could you explain what's happening a bit more in detail ?

They way we decide whether to call PostBeginPlay is by hooking the regular AActor::BeginPlay function and checking if the name of the actor exists as a mod, and if it does then we call PostBeginPlay after AActor::BeginPlay is done executing.
So the PostBeginPlay function for each mod is supposed to be called every time that mod actor is created, which should only be once every time a map loads.

For PreBeginPlay we just call it once every time a BP mod is loaded by the loader, which is also supposed to be just once every time a map loads.

Issue #91, opened 2023-03-24
[BUG - Dev] Post begin play BP event firing twice per BP


narknon:
**Branch or Release**
Current Dev

**Game and Engine Version**
4.27, likely others

**Describe the bug**
Postbeginplay custom event hook fires twice.

**To Reproduce**
1. Create BP mod with postbeginplay event
2. Launch game, loading bp mod with ue4ss
3. See duplicative call in log

**Expected behavior**
Call once.

**Desktop (please complete the following information):**
 - OS: Win11

**Additional context**

narknon:
Seems to have gone away most of the time..

PR #447, opened 2024-03-29
Fixed BPModLoaderMod RegisterBeginPlayPostHook

### LINKED ISSUE #433: [BUG - Release] BPModLoaderMod call PostBeginPlay() for every logic mod as many times as there is logic mods
**Release**
v3.0.1

**Game and Engine Version**
Found on the following, but affects every game which use this version of BPModLoaderMod.
Palworld 0.1.5.1 Steam
UE5.1.1 as far as I know

**Describe the bug**
BPModLoaderMod spamming log with the same messages "PostBeginPlay not valid for mod" and "Executing 'PostBeginPlay' for mod".
And it seems to call PostBeginPlay() as many times for every logic mod as there's logic mods installed.
Most likely due to a misplaced end of the loop. Please refer to log files attached below.

Proposed fix:
* Current code
```
RegisterBeginPlayPostHook(function(ContextParam)
    local Context = ContextParam:get()
    for _,ModConfig in ipairs(OrderedMods) do
        if Context:GetClass():GetFName() ~= ModConfig.AssetNameAsFName then return end
        local PostBeginPlay = Context.PostBeginPlay
        if PostBeginPlay:IsValid() then
            Log(string.format("Executing 'PostBeginPlay' for mod '%s'\n", Context:GetFullName()))
            PostBeginPlay()
        else
            Log(string.format("PostBeginPlay not valid for mod %s\n", Context:GetFullName(), true))
        end
    end
end)
```
add another `end` between `if Context:GetClass():GetFName() ~= ModConfig.AssetNameAsFName then return end`
and `local PostBeginPlay = Context.PostBeginPlay`. Remove second `end` from the end.

* Fixed version
```
RegisterBeginPlayPostHook(function(ContextParam)
    local Context = ContextParam:get()
    for _,ModConfig in ipairs(OrderedMods) do
        if Context:GetClass():GetFName() ~= ModConfig.AssetNameAsFName then return end
    end

    local PostBeginPlay = Context.PostBeginPlay
    if PostBeginPlay:IsValid() then
        Log(string.format("Executing 'PostBeginPlay' for mod '%s'\n", Context:GetFullName()))
        PostBeginPlay()
    else
        Log(string.format("PostBeginPlay not valid for mod %s\n", Context:GetFullName(), true))
    end
end)
```

Considering ModConfig.AssetNameAsFName seem to be the same for all logic mods that for loop seem to work fine, but I'm not sure this will work right in case these mods end up with different FNames if that's possible since it may skip contexts it isn't supposed to skip. Also, I'm not sure why it ever reaches PostBeginPlay() considering strings Context:GetClass():GetFName() and ModConfig.AssetNameAsFName seem to never match. I've logged them (with %s) and they are alwasy different, but ~= somehow returns 'false' like they are the same when one of the logic mods is the context.

**Mods directory**
Not necessary, doesn't depend on script mods or specific content of logic mods. I can attach all logic mods I use if required, though.

**To Reproduce**
0. Add more than 1 logic mod. Preferably 3 or more.
1. Launch the game
2. Exit the game
3. Look at the log file

**Expected behavior**
BPModLoaderMod calls PostBeginPlay() only once per logic mod installed. From my experience that's enough for logic mods to work. At least for me.

Example of a log with repeated messages and calls. For example, note how InfiniteWeightInCamp mod reports that it loaded correctly multiple times.
[UE4SS-bug.log](https://github.com/UE4SS-RE/RE-UE4SS/files/14552912/UE4SS-bug.log)

Log with fixed loop when each mod executed only once. 
[UE4SS-fixed.log](https://github.com/UE4SS-RE/RE-UE4SS/files/14552913/UE4SS-fixed.log)

Same as above, but I loaded my save and then exit the game.
[UE4SS.log](https://github.com/UE4SS-RE/RE-UE4SS/files/14553008/UE4SS.log)

The entire mod loding still seem to happen twice every time it happen. Probably there's something else wrong, but I have no idea what.

 - OS: Win10

**Additional context**

Okaetsu:
**Description**

I noticed we're only checking for the class name `ModActor_C` inside `RegisterBeginPlayPostHook` rather than the full class path so I added a check against the full class path which will only trigger `PostBeginPlay` once for the ModActor that fired BeginPlay.

The boolean for `Log` was accidentally placed inside string.format so it'd still be printing `PostBeginPlay not valid` with `VerboseLogging` set to false.

Fixes #433, #91

**Type of change**

**How Has This Been Tested?**

Tested on 4.27 and 5.1 with multiple ModActors.

**Checklist**

### DIFF
--- a/assets/Changelog.md
+++ b/assets/Changelog.md
@@ -39,6 +39,10 @@ Added support for watching ArrayProperty and StructProperty.
 ### General
 Fixed BPModLoaderMod giving "bad conversion" errors.
 
+Fixed BPModLoaderMod calling `PostBeginPlay` multiple times for each ModActor
+
+Fixed BPModLoaderMod displaying 'PostBeginPlay not valid' when VerboseLogging is set to false
+
 Fixed some debug GUI layout alignments, especially with different GUI font scaling settings.
 
 ### Live View
--- a/assets/Mods/BPModLoaderMod/Scripts/main.lua
+++ b/assets/Mods/BPModLoaderMod/Scripts/main.lua
@@ -266,7 +266,7 @@ RegisterBeginPlayPostHook(function(ContextParam)
     local Context = ContextParam:get()
     for _,ModConfig in ipairs(OrderedMods) do
         if Context:GetClass():GetFName() ~= ModConfig.AssetNameAsFName then return end
-        local AssetPathWithClassPrefix = "BlueprintGeneratedClass " .. ModConfig.AssetPath .. "." .. ModConfig.AssetNameAsFName:ToString()
+        local AssetPathWithClassPrefix = "BlueprintGeneratedClass " .. ModConfig.AssetPath .. "." .. ModConfig.AssetName
         if AssetPathWithClassPrefix == Context:GetClass():GetFullName() then
             local PostBeginPlay = Context.PostBeginPlay
             if PostBeginPlay:IsValid() then
--- a/assets/Mods/BPModLoaderMod/Scripts/main.lua
+++ b/assets/Mods/BPModLoaderMod/Scripts/main.lua
@@ -266,12 +266,15 @@ RegisterBeginPlayPostHook(function(ContextParam)
     local Context = ContextParam:get()
     for _,ModConfig in ipairs(OrderedMods) do
         if Context:GetClass():GetFName() ~= ModConfig.AssetNameAsFName then return end
-        local PostBeginPlay = Context.PostBeginPlay
-        if PostBeginPlay:IsValid() then
-            Log(string.format("Executing 'PostBeginPlay' for mod '%s'\n", Context:GetFullName()))
-            PostBeginPlay()
-        else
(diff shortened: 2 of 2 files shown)


UE4SS:
Seems this should fix #433.
It might also at least partially fix #370, and I don't understand how the PreBeginPlay could possibly be duplicated, maybe a misunderstanding of some sort.
#91 was probably also this same problem.

UE4SS (approved):
The code changes look good, but beware: I haven't tested it.

Okaetsu:
> Seems this should fix #433. It might also at least partially fix #370, and I don't understand how the PreBeginPlay could possibly be duplicated, maybe a misunderstanding of some sort. #91 was probably also this same problem.

I'm not certain what the issue with PreBeginPlay would be as it is working perfectly fine for me and only fires once per ModActor. It does sound like a misunderstanding to me.

Buckminsterfullerene02:
Works fine for me.

PR #463, opened 2024-04-08
fix: compilation quirk with vs2019

Lyrth:
### Description

It seems that VS2019 should still be supported according to the readme. This should work around the issue with building with (some versions of?) VS2019, where [LuaMod.cpp#L2206](https://github.com/UE4SS-RE/RE-UE4SS/blob/73d337852d004f71bd1b5d24ad703263854e44e7/UE4SS/src/Mod/LuaMod.cpp#L2206) fails with "`error C3493: 'i' cannot be implicitly captured because no default capture mode has been specified`," which I assume is a MSVC quirk, as it doesn't make sense otherwise. Happened on a windows-2019 Github runner, and coincidentally on one other person too.

### Type of change

### How Has This Been Tested?

<details>
<summary>Testing</summary>

Test code:
```lua
    local dirs = IterateGameDirectories()
    for k, GameDirectory in pairs(dirs) do
        print(("%s: %s [%s]"):format(k, GameDirectory.__name, GameDirectory.__absolute_path))
        for i, File in pairs(GameDirectory.__files) do
            print(("    - %s: %s [%s]"):format(i, File.__name, File.__absolute_path))
        end
    end
end)
```

Directory layout:
image

Output (Release [UE4SS - v3.0.1 Beta #0 - Git SHA #d935b5b]):
image

Output (this change):
image

</details>

### Checklist

### Additional context

Unsure about appending to changelog again, because the xmake change might or *might not* be involved in this, given that the error has not appeared before for anyone it seems, except recently (today).

### DIFF
--- a/UE4SS/src/Mod/LuaMod.cpp
+++ b/UE4SS/src/Mod/LuaMod.cpp
@@ -2199,17 +2199,18 @@ Overloads:
                                     }
                                     const auto current_path = std::string{lua.get_string()};
                                     auto files_table = lua.prepare_new_table();
-                                    for (int i = 1; const auto& item : std::filesystem::directory_iterator(current_path))
+                                    auto index = 1;
+                                    for (const auto& item : std::filesystem::directory_iterator(current_path))
                                     {
                                         if (!item.is_directory())
                                         {
-                                            files_table.add_key(i);
+                                            files_table.add_key(index);
                                             auto file_table = lua.prepare_new_table();
                                             file_table.add_pair("__name", to_string(item.path().filename().wstring()).c_str());
                                             file_table.add_pair("__absolute_path", to_string(item.path().wstring()).c_str());
                                             files_table.fuse_pair();
                                         }
-                                        ++i;
+                                        ++index;
                                     }
                                     return 1;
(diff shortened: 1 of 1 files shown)


Lyrth:
[Relevant GitHub workflow run and line](https://github.com/Lyrth/RE-UE4SS/actions/runs/8597772149/job/23557158154#step:7:358). There's not much context with it aside from some unrelated warnings.

UE4SS (approved):
I'm not entirely sure why this all of a sudden was a problem but the fix looks good.

As far as the changelog is concerned, since this is a recent problem (appeared after the current release), adding it to the changelog would be nonsensical because users updating from the current release to the next release will never have experienced this problem.

bitonality:
Can confirm that I was able to reproduce the problem locally by running `xmake f --vs=2019` to explicitly set the xmake VS version to VS2019 16.11.33. Seems like it's been a problem that has regressed occasionally with the 2019 toolchain ([MSFT Bug Report](https://developercommunity.visualstudio.com/t/c3493-with-stdclatest-using-structured-binding-in/1318655)).

PR #464, opened 2024-04-09
Add optimization and symbols for xmake build

### LINKED ISSUE #462: Optimization flags wrong & no symbols with xmake
The xmake build system isn't properly setting optimization flags.

The optimization flags (/O2, /O0d) aren't being set for any of the configs.
The PDB files aren't being generated properly either, which I noticed when I tried to debug a problem and VS was unable to find any source information in the PDB.

It seems to me that the PDB problem can be solved by adding `["symbols"] = {"debug"}` to all the configs instead of just the shipping config.
The optimization flags problem is not one that I know how to fix.

I haven't made a PR for the PDB problem.
I suspect it will involve working on the same file as the optimization fix and I don't feel like dealing with the conflict so if whoever works on the optimization fix doesn't include the PDB fix as well then I'll make a PR after the optimization fix has been merged.

bitonality:
**Description**

Fixes #462 

**Type of change**

Bug fix (non-breaking change which fixes an issue)

**How Has This Been Tested?**

Tested with `xmake` from cmd and tested with VS2022 generated project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,13 @@ Afterwards open the generated `.sln` file inside of the `vsxmake2022` directory
 
 Note that you should also commit & push the submodules that you've updated if the reason why you updated was not because someone else pushed an update, and you're just catching up to it.
 
+**Note the following about how xmake interacts with VS**
+
+> The vs. build plugin performs the compile operation by directly calling the xmake command under vs, and also supports intellisense and definition jumps, as well as breakpoint debugging.
+
+This means that modifying the project properties within Visual Studio will not affect which flags are passed to the build when VS executes `xmake`. XMake provides some configurable project settings
+which can be found in VS under the `Project Properties -> Configuration Properties -> Xmake` menu.
+
 ## Updating git submodules
 
 If you want to update git submodules, you do so one of three ways:
--- a/tools/xmakescripts/build_configs.lua
+++ b/tools/xmakescripts/build_configs.lua
@@ -19,27 +19,34 @@ TARGET_TYPES = {
 
 CONFIG_TYPES = {
     ["Dev"] = {
+        ["symbols"] = {"debug"},
         ["defines"] = {
             "UE_BUILD_DEVELOPMENT",
             "STATS"
-        }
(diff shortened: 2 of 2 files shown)


UE4SS:
Using  `xmake project -k vsxmake2022 -y`, VS displays `/O2` with all build modes for all targets with this patch.
I tested both with the mod set up and standalone, same results.

2024-04-10_00-37_wrong_opt_level

bitonality:
Interesting... It looks like xmake appends the correct optimization flags at build time instead of setting them in the vs project.

When I run a verbose build, I can see that the correct flags are being set in the logs.
Optimization

When I manually try and change the optimization setting from within VS, it adds the following to the vsproject file
```xml
  <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='CasePreserving__Dev__Win64|x64'">
    <ClCompile>
      <Optimization>MinSpace</Optimization>
    </ClCompile>
  </ItemDefinitionGroup>
  ```
  
 Which will likely get trampled by xmake upon a `xmake project -k vsxmake2022`.
 
 So this would appear to be a side effect of how xmake integrates with VS project files and not a failure in functionality.

Thoughts?

UE4SS:
> Thoughts?

Looks like xmake with VS isn't actually using VS to build but simply executes xmakes build command.
It also appears to execute xmakes config command every time before building.
As such, the settings in VS probably don't matter at all since it'll use whatever was configured with xmake.
The size differences in UE4SS.dll between the shipping and debug builds reinforces this as well.
We should probably warn about this as part of the build instructions.

narknon:
Are you saying you can't change the build target in VS?

UE4SS:
> Are you saying you can't change the build target in VS?

No I'm saying everything is fine except I suspect that project properties don't apply (xmake overrides them) so we should warn about this in the build instructions so that people don't make changes there and expect it to work.
Changing the build target works fine.

UE4SS (approved):
This look good to me.
Tested and it works.

PR #472, opened 2024-04-17
Live view search improvement

Buckminsterfullerene02:
**Description**

Feels good to be able to contribute a little bit of code for things that have been bugging me for a while now that I can build UE4SS.

- Changes ExcludeClassNames filter to a dropdown to change between IncludeClassNames and ExcludeClassNames. This is because they both cannot be used at once.
- IncludeClassNames (which could be renamed to IncludeOnlyClassNames for clarity) does the opposite of ExcludeClassNames.
- ClassNameFilter, HasProperty and HasPropertyType filters now all support comma seperated lists.
- Save filters to file, which is currently stored in `liveview/filters.meta.json`. I'm not sure about this location as if we keep going this way we could end up with loads of loose folders in the install directory and thats not great. Ideally we could just have one folder called configs or similar, and inside that have filters, watches and whatever other subfolders for runtime config files. 
- Refresh search button. Perhaps slightly jank because it would be ideal to just re-search for the one input settings (i.e. all checkboxes other than function parameter flags), but this is a lot nicer UX than clicking back on search bar and hitting enter for the more than one input settings.

**Type of change**

**How Has This Been Tested?**

Ran DRG and tested various user input scenarios and added extra validation based on those testing rounds.

**Checklist**

image

Video showing filters: https://drive.google.com/file/d/1H3mepYLvtrLQAGyQ0GLORFQK-UBZJDfV/view?usp=sharing

**Additional context**

Todo:
- [x] Fix loading from filters file, something to do with LoopAction but I don't understand what I'm doing wrong with it. The lambda seems to expect a LoopAction return which I've done but no luck. When this is fixed loading filters from file can be tested. @UE4SS any help? Specifically [this area](https://github.com/UE4SS-RE/RE-UE4SS/pull/472/files#diff-90fa070ccf3f3e996aa2d423a8eaeb939db54d673f4d9534823e6247cd4ef22dR351-R405)
- [x] Docs
- [x] Changelog
- [x] Run clang-format
- [x] Squash before merge


UE4SS:
Regarding the error with LoopAction, you're missing returns in these three lambdas, they all should return `LoopAction::Continue`:
2024-04-17_14-46_lambdas_missing_return

UE4SS:
Added `Run clang-format` to the todo list as it appears to not have been run.

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -279,6 +280,137 @@ namespace RC::GUI
     };
     FLiveViewDeleteListener FLiveViewDeleteListener::LiveViewDeleteListener{};
 
+    static auto add_bool_filter_to_json(JSON::Array& json_filters, const StringType& filter_name, bool is_enabled) -> void {
+        auto json_object = std::make_unique<JSON::Object>();
+        json_object->new_string(STR("FilterName"), filter_name);
+        auto& filter_data = json_object->new_object(STR("FilterData"));
+        filter_data.new_bool(STR("Enabled"), is_enabled);
+        json_filters.add_object(std::move(json_object));
```
UE4SS, line 288:
```suggestion
        auto& json_object = json_filters.new_object();
        json_object.new_string(STR("FilterName"), filter_name);
        auto& filter_data = json_object.new_object(STR("FilterData"));
        filter_data.new_bool(STR("Enabled"), is_enabled);
```
While this code should function, there is a helper function that avoids explicit memory allocation, dealing with a pointer, and std::move.
Note that I'm not necessarily asking you to change this, because I realize you probably copied it from the watches file code, but I do want you to be aware of this API still.
Feel free to resolve this with the button without applying any changes.

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -279,6 +280,137 @@ namespace RC::GUI
...
+            if constexpr (std::is_same_v<typename ContainerType::value_type, FName>) {
+                StringType string_value = value.ToString();
+                values_array.new_string(string_value);
+            }
+            else
+            {
+                values_array.new_string(value);
+            }
+        }
+        json_filters.add_object(std::move(json_object));
```
UE4SS, line 313:
```suggestion
        auto& json_object = json_filters.new_object();
        json_object.new_string(STR("FilterName"), filter_name);
        auto& filter_data = json_object.new_object(STR("FilterData"));

        if (array_name == STR("ClassNames"))
        {
            filter_data.new_bool(STR("IsExclude"), Filter::ClassNamesFilter::b_is_exclude);
        }

        auto& values_array = filter_data.new_array(array_name);
        for (const auto& value : container) {
            if constexpr (std::is_same_v<typename ContainerType::value_type, FName>) {
                StringType string_value = value.ToString();
                values_array.new_string(string_value);
            }
            else
            {
                values_array.new_string(value);
            }
        }
```
This is another use of unique_ptr that can be removed, it's exactly the same as the earlier one.
Feel free to resolve this with the button without applying any changes.

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -279,6 +280,137 @@ namespace RC::GUI
...
+        if (array_name == STR("ClassNames"))
+        {
+            filter_data.new_bool(STR("IsExclude"), Filter::ClassNamesFilter::b_is_exclude);
+        }
+
+        auto& values_array = filter_data.new_array(array_name);
+        for (const auto& value : container) {
+            if constexpr (std::is_same_v<typename ContainerType::value_type, FName>) {
+                StringType string_value = value.ToString();
+                values_array.new_string(string_value);
```
UE4SS, line 306:
```suggestion
                values_array.new_string(value.ToString());
```
I think this is an unnecessary temporary variable ?
It doesn't generate a compiler error.

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -279,6 +280,137 @@ namespace RC::GUI
...
+        );
+        auto json_file_contents = json_file.read_all();
+        if (json_file_contents.empty()) {
+            return;
+        }
+
+        const auto json_global_object = JSON::Parser::parse(json_file_contents);
+        const auto& json_filters = json_global_object->get<JSON::Array>(STR("Filters"));
+        json_filters.for_each([&](const JSON::Value& filter) -> LoopAction {
+            auto& json_object = *filter.as<JSON::Object>();
```
UE4SS, line 366:
```suggestion
            if (!filter.is<JSON::Object>()) {
                throw std::runtime_error{"Invalid filter in 'filters.meta.json'"};
            }
            auto& json_object = *filter.as<JSON::Object>();
```
Again not asking you to make any changes since you've probably copied the code for watches, I'm just making you aware of how the API works.

The `as<T>` function makes an assumption that you've already confirmed that the type stored is actually `T`.
Your code doesn't confirm this.
It will still work but it will crash if this file is edited (externally from UE4SS) and the object is no longer an object.
Manual user edits or other external edits aside, what happens if we change the format of this file later ?
It will crash because we assumed T == Object but in the new format it might not be an object.

The `get<T>` function also works the same way.

My code suggestion shows how you would confirm the type.
I chose this particular bit of code for this example but there's plenty of places where this happens, just fyi.

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -2950,31 +3081,123 @@ namespace RC::GUI
...
+                        }
+                        else
+                        {
+                            Output::send(STR("Search for: {}\n"), search_buffer.empty() ? STR("") : to_wstring(search_buffer));
+                            s_name_to_search_by = search_buffer;
+                            m_object_iterator = &LiveView::guobjectarray_by_name_iterator;
+                            m_is_searching_by_name = true;
+                            search_by_name();
+                        }
+                    }
```
UE4SS, line 3189: This is code duplication.
Can we make this into a member function that we can use here and at the original location ?

UE4SS (changes requested):
This looks good for the most part.
There's one major problem, and then a few lesser problems, including some that appear to be bad choices by me in the original code that some of this appears to be based on.

UE4SS/include/GUI/LiveView/Filter/HasPropertyType.hpp
```diff
@@ -1,37 +1,40 @@
...
             for (const auto& property : as_struct->ForEachPropertyInChain())
             {
-                if (property->GetClass().GetFName() == s_value)
+                for (const auto& property_type : list_property_types)
                 {
-                    should_filter_object_out = false;
-                    break;
+                    if (property->GetClass().GetFName() == property_type)
+                    {
+                        return false;
```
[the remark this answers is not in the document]
Buckminsterfullerene02: Rookie mistake!

UE4SS/src/GUI/LiveView.cpp
Buckminsterfullerene02: Just didn't think about embedding it into the same line when adding it.

UE4SS/src/GUI/LiveView.cpp
Buckminsterfullerene02: Thanks for explaining, I understand that now.

Buckminsterfullerene02:
@UE4SS There is an issue with reading the json file
image

Error message suggests that at line 7 column 1 in the file there is a character instead of a `"` but it's an indentation 
image

I did some debugging but I don't have the time to try solving it myself given I know nothing about your json library so I'll just let you do it. 

Latest commit has the save to filter button temporarily running loading from the file (so I could debug it by attaching to process after load) but you'll probably want to change that to save then load or whatever you do.

> when someone has the time and willpower the change can be made for existing files.
It's a low priority thing so when I have more than an hour a day I can take a look at it, for a different PR. Definately agree about having the settings moved into there, with backwards compatibility of course. But would have to bring it up with others to see what they think about it before doing anything.

UE4SS:
When did this error start happening ?
After you made some changes ?

Buckminsterfullerene02:
> When did this error start happening ?
> After you made some changes ?

As soon as I fixed the code that loads the file from disk so it could build successfully, this error message appeared.

UE4SS:
Try it now.
Also while testing this fix, I noticed you haven't added all the saved filters to the code that loads them from file.

Buckminsterfullerene02:
> Try it now. Also while testing this fix, I noticed you haven't added all the saved filters to the code that loads them from file.

Didn't seem to fix it completely. Check now with it trying to load the function param flag checkbox values.

UE4SS:
That should be fixed now.

While I was at it, I noticed that the function param flags are not loading properly.
This is because you're setting the checkbox for all flags to (effectively) the last entry in the saved file.
Replace the loop with something like this to fix it:
```c++
function_param_flags.for_each([](auto index, JSON::Value& flag) {
    // ... error checking removed for brevity, obviously keep that.

    // The order that the flags are saved to file (and loaded) MUST match their entries into 's_checkboxes'.
    Filter::FunctionParamFlags::s_checkboxes[index] = flag.as<JSON::Bool>()->get();
    return LoopAction::Continue;
}
```

UE4SS:
Also, you may want to make sure that `function_param_flags.get().size() == Filter::FunctionParamFlags::s_checkboxes.size()` just in case the file has too few or too many flags somehow.

UE4SS (approved):
The code looks good to me.

PR #478, opened 2024-04-21
chore: Moved changelog entries to their proper places

UE4SS:
**Type of change**

- [x] Other... Please describe: Changelog update


Buckminsterfullerene02:
Was thinking of this too... while you're at it, maybe make a new subsection for BPModLoader?

UE4SS:
> Was thinking of this too... while you're at it, maybe make a new subsection for BPModLoader?

Sure.

But if we're doing more changes, perhaps we should reconsider empty sub-sections ?
I looked at the changelog recently and realized it could actually be confusing when you have an empty sub-section right before another section.

For example:

## New

### General

Cool stuff.

### Experimental

## Changes

### General

Cool change.

The "Cool change" could be misinterpreted as an experimental change.
I think we should either
A: Put "None" in each empty sub-section, to be replaced by whoever is first to make a change appropriate for that sub-section.
or
B: Remove empty sections immediately before release, and re-add them immediately after release.
That way things are still where users expect them to be and the ease of knowing exactly where to add entries also remains.

Thoughts ?

Buckminsterfullerene02:
Remove empty subsections

UE4SS:
> Remove empty subsections just before release

Alright, I'll assume you'll add that to the checklist on the discord.

Buckminsterfullerene02 (approved):

UE4SS:
> conflict resolution bump

I believe this PR is ready to merge.

PR #481, opened 2024-04-21
fix(C++ Mods): Delayed mod constructor until imgui is initialized

UE4SS:
**Description**

This only happens if the mod uses `UE4SS_ENABLE_IMGU`' macro in the constructor.

The purpose of this is to guarantee that imgui context has been set before a mod uses any imgui functions. Previously, we were assuming that the context would be set by the time the mod construct was called, however this was actually a race condition. It was possible for the mod constructor to be called before the UI code in UE4SS created the imgui context.

**Type of change**

**How Has This Been Tested?**

See #469

**Checklist**

### DIFF
--- a/UE4SS/include/Mod/CppMod.hpp
+++ b/UE4SS/include/Mod/CppMod.hpp
@@ -60,6 +60,7 @@ namespace RC
                 -> void;
 
         auto fire_unreal_init() -> void override;
+        auto fire_ui_init() -> void override;
         auto fire_program_start() -> void override;
         auto fire_update() -> void override;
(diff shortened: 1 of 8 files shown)


UE4SS:
I've realized this fix isn't sufficient.
It will cause loading to stop if the GUI isn't enabled & visible on startup.
Will fix this soon.

UE4SS:
The proper fix will probably involve introducing `on_ui_init` for C++ mods.
That will be the place where a C++ mod uses the `UE4SS_ENABLE_IMGUI` macro.

UE4SS:
This is now ready.

narknon (approved):

PR #488, opened 2024-05-02
chore: update changelog

localcc:


narknon (approved):
Please mention this will be a breaking change if default is currently being used.

PR #506, opened 2024-05-15
New UE4SS working directory location

Buckminsterfullerene02:
**Description**

As according to the discussions in #476. Install location of all files will be in <game root dir>/ue4ss except proxy. Legacy locations all supported, but if files are in new locations, those take priority.

Proxy tries to load UE4SS.dll from new location, but the location of the UE4SS.dll (and thus the root directory) can be overriden by placing a file called `override.txt` next to the proxy and can contain either an absolute or relative path to the UE4SS.dll. Still supports legacy UE4SS.dll location.

Also changes the build script as all files need to be zipped inside of a `ue4ss/` folder that will still be unzipped inside of the game exe directory due to the proxy also being there.

**Type of change**

**How Has This Been Tested?**

Tested files in legacy, new, and overriden file locations, seperately and in a mix of two or three at once.

**Checklist**

- [x] Fix issue with console/dll not initialising game correctly
- [x] Fix release assets not going into the right places


Buckminsterfullerene02:
If running the game directly from the shipping exe, there is an issue where DRG doesn't load properly as shown:

image

If running the game from steam, it doesn't load up at all.

I'm sure both of these issues are related to the proxy dll, but I wouldn't even know where to begin with debugging as there is nothing obvious when breakingpoint debugging the proxy code.

UE4SS:
> If running the game directly from the shipping exe, there is an issue where DRG doesn't load properly as shown:
> 
> If running the game from steam, it doesn't load up at all.
> 
> I'm sure both of these issues are related to the proxy dll, but I wouldn't even know where to begin with debugging as there is nothing obvious when breakingpoint debugging the proxy code.

As far as Steam not working, that's because you're using `std::filesystem::current_path()` which gives you the working directory.
Steam actually uses the location of the non-shipping exe as the working directory.
I've implemented the solution to this by using `GetModuleFileNameW` instead.

As far as DRG not loading properly when using the shipping exe, what do you mean ?
I don't see anything odd other than the regular Windows console not rendering properly for whatever reason.
I was unable to reproduce that problem.

UE4SS:
> I'm sure both of these issues are related to the proxy dll, but I wouldn't even know where to begin with debugging as there is nothing obvious when breakingpoint debugging the proxy code.

To debug the proxy, open `Intermediates/.gens/proxy_files/windows/x64/Game__Debug__Win64/dllmain.cpp` in your IDE, put a breakpoint, and then set your IDE up to launch the game and attach when you hit the debug button, and then your breakpoint should be hit immediately.
Replace `Game__Debug__Win64` in the path with whatever target you built.
I figured this problem out by putting a breakpoint in that file in the load_ue4ss_dll function and inspecting the value of `ue4ssPath`.

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -581,12 +581,26 @@ namespace RC::GUI
...
+            file_opened = true;
+        }
+        else if (std::filesystem::exists(legacy_root_directory_path))
+        {
+            auto json_file = File::open(legacy_root_directory_path, File::OpenFor::Reading, File::OverwriteExistingFile::No, File::CreateIfNonExistent::Yes);
+            json_file_contents = json_file.read_all();
+            file_opened = true;
+        }
+        
+        if (!file_opened || json_file_contents.empty())
```
UE4SS, line 603:
```suggestion
        bool is_legacy = !std::filesystem::exists(working_directory_path) && std::filesystem::exists(legacy_root_directory_path);
        auto json_file = File::open(is_legacy ? legacy_root_directory_path : working_directory_path, File::OpenFor::Reading, File::OverwriteExistingFile::No, File::CreateIfNonExistent::Yes);
        
        if (json_file_contents.empty())
```
This can be simplified.

UE4SS/proxy_generator/main.cpp
```diff
@@ -176,13 +181,73 @@ int _tmain(int argc, TCHAR* argv[])
...
-    cpp_file << "        setup_functions();" << endl;
-    cpp_file << "        LoadLibrary(STR(\"UE4SS.dll\"));" << endl;
+    cpp_file << "        HMODULE hUE4SSDll = load_ue4ss_dll(hInstDll);" << endl;
+    cpp_file << "        if (hUE4SSDll)" << endl;
+    cpp_file << "        {" << endl;
+    cpp_file << "            setup_functions();" << endl;
+    cpp_file << "        }" << endl;
+    cpp_file << "        else" << endl;
+    cpp_file << "        {" << endl;
+    cpp_file << "            ExitProcess(0);" << endl;
```
UE4SS, line 249: Should we put a MessageBox here maybe to explain what happened to the user ?

UE4SS:
Packaging is failing: https://github.com/UE4SS-RE/RE-UE4SS/actions/runs/9106272884/job/25033211418

Buckminsterfullerene02:
> > I'm sure both of these issues are related to the proxy dll, but I wouldn't even know where to begin with debugging as there is nothing obvious when breakingpoint debugging the proxy code.
> 
> To debug the proxy, open `Intermediates/.gens/proxy_files/windows/x64/Game__Debug__Win64/dllmain.cpp` in your IDE, put a breakpoint, and then set your IDE up to launch the game and attach when you hit the debug button, and then your breakpoint should be hit immediately. Replace `Game__Debug__Win64` in the path with whatever target you built. I figured this problem out by putting a breakpoint in that file in the load_ue4ss_dll function and inspecting the value of `ue4ssPath`.

That's what I was doing, I just didn't see anything wrong, so I thought it was some weird thing I didn't understand.

UE4SS/proxy_generator/main.cpp
Buckminsterfullerene02: I put in two; one for when proxy fails to load for whatever reason, and one for when UE4SS.dll loads. Not sure about the error message, currently I just have "Failed to load UE4SS.dll" but don't know if it should be any more informative. Like I need to write up docs on the override file and such, so maybe the message can point to read that to make sure it's installed in valid locations.

UE4SS: Yeah that might be good.

narknon:
There shouldn't be a message box when ue4ss loads should there? That's just kinda annoying if you have to close out of it

UE4SS:
> There shouldn't be a message box when ue4ss loads should there? That's just kinda annoying if you have to close out of it

It's only if it fails to load.
Otherwise the game just immediately shuts down without an error, which means the user has no idea what happened.

UE4SS/proxy_generator/main.cpp
UE4SS: We should put some kind of prefix or title or something on these message boxes that tells users that this message is from UE4SS.

narknon:
"I put in two; one for when proxy fails to load for whatever reason, and one for when UE4SS.dll loads."

UE4SS:
> "I put in two; one for when proxy fails to load for whatever reason, and one for when UE4SS.dll loads."

That's a typo, they meant if UE4SS.dll fails to load.

UE4SS:
> The second half of the sentence

Yes, `and one for when UE4SS.dll loads` is a typo.
I literally checked the code as well.

tools/buildscripts/build.py
```diff
@@ -160,36 +160,48 @@ def package_release(is_dev_release: bool):
...
+        # Proxy
         shutil.copy(dwmapi_dll_path, staging_dir)
 
         if is_dev_release:
-            shutil.copy(ue4ss_pdb_path, staging_dir)
-            if os.path.exists(os.path.join(staging_dir, 'docs')):
-                shutil.copytree('docs', os.path.join(staging_dir, 'docs'))
+            shutil.copy(ue4ss_pdb_path, ue4ss_dir)
+            if os.path.exists(os.path.join(scan_start_dir, 'docs')):
+                shutil.copytree('docs', os.path.join(ue4ss_dir, 'Docs'))
```
UE4SS, line 185: Is this capitalization an erroneous change ?

Buckminsterfullerene02: No, all the other folders are capitalized and it just looks out of place not being so, and I don't see any harm in it.

UE4SS: Two things:
1. Will this work on Linux ? Because names are case-sensitive on Linux.
2. The name was only changed in the call to copy the directory, not the call above to see if the directory exists, this alongside the fact that the real directory is all lowercase is what made this stand out to me as a potential error.

Buckminsterfullerene02: If checks if the docs directory exists (it probably doesn't need to since we know it does, but better safe than sorry), then copies the docs directory to the new location and changes the name to have a capital D. I have no reason to suspect this wouldn't work on Linux, though the build scripts may need to be updated to work on Linux when the Linux port gets to that point.

UE4SS: Okay I understand what it does now.
I misunderstood earlier.
This potential problem is a non-problem, no further discussion is needed.
I did test `os.path.exists` just out of curiosity, and it fails if the casing is wrong on Linux, just for future reference.

Buckminsterfullerene02:
Just realised deleting API.txt broke build script. Will get on that shortly.

Buckminsterfullerene02:
Releases build as expected now.

UE4SS:
Does this PR also affect the location of imgui.ini ? Not sure if the creation of that file is based on the location of UE4SS.dll or the exe.

Buckminsterfullerene02:
> Does this PR also affect the location of imgui.ini ? Not sure if the creation of that file is based on the location of UE4SS.dll or the exe.

Forgot about that. It seems to still be put into the exe dir.

UE4SS:
> > Does this PR also affect the location of imgui.ini ? Not sure if the creation of that file is based on the location of UE4SS.dll or the exe.
> 
> Forgot about that. It seems to still be put into the exe dir. Should it be kept that way?

If possible, it should go in the ue4ss directory, but I don't know if imgui provides a way to do that and I don't think we should bend over backwards to get that to work.

UE4SS:
> Revert moving imgui.ini location
> Don't ask me why, I have literally no idea.

What ? Was there something that didn't work properly ?

UE4SS:
The object dumper directory doesn't use the new path:
```
[15:19:47] mods directory: Y:\test_games\5.4\UE4SS_Main54\Windows\UE4SS_Main54\Binaries\Win64\ue4ss\Mods
[15:19:47] log directory: Y:\test_games\5.4\UE4SS_Main54\Windows\UE4SS_Main54\Binaries\Win64\ue4ss
[15:19:47] object dumper directory: Y:\test_games\5.4\UE4SS_Main54\Windows\UE4SS_Main54\Binaries\Win64
```

UE4SS (approved):
Looks good to me.
Tested and it's working.

PR #508, opened 2024-05-16
feat: build caching

Buckminsterfullerene02:
**Description**

Adds github action caching to the release actions. Currently it only works for xmake dependency downloads, so cuts down about 30% of the build time. Could not get it to work for incremental compiling though (tried adding `--ccache=y` to build target command, and adding `build.cache` policy in `xmake.lua`).

**Type of change**

- [x] Other... Github action

**How Has This Been Tested?**

Ran experimental action on branch a few times to check that the caching is being done properly.

**Checklist**

N/A

### DIFF
--- a/.github/workflows/experimental.yml
+++ b/.github/workflows/experimental.yml
@@ -41,6 +41,16 @@ jobs:
         with:
           xmake-version: latest
 
+      - name: Cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            .xmake
+            Binaries
+            Intermediates
+            C:/Users/runneradmin/AppData/Local/.xmake
+          key: ${{ runner.os }}-xmake-${{ hashFiles('**/xmake.lua') }}
+
       - name: Build
         run: |
           xmake f -m "Game__Shipping__Win64" -y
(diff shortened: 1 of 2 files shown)


bitonality:
> (tried adding `--ccache=y` to build target command, and adding `build.cache` policy in `xmake.lua).

the ccache option should be part of configuration and not build command - i.e. `xmake f -m "Game__Shipping__Win64" --ccache=y`

Not sure if that's what you meant, let me know.

bitonality:
Also dumb question,

Is ccache installed on the runner?

Buckminsterfullerene02:
> > (tried adding `--ccache=y` to build target command, and adding `build.cache` policy in `xmake.lua).
> 
> the ccache option should be part of configuration and not build command - i.e. `xmake f -m "Game__Shipping__Win64" --ccache=y`
> 
> Not sure if that's what you meant, let me know.

Yeah, that's what I meant when I said "build target command".

Buckminsterfullerene02:
> Also dumb question,
> 
> Is ccache installed on the runner?

I just assumed it was part of xmake, so I don't know...

Buckminsterfullerene02:
Turns out the `build.cache` policy really messes up proxy gen so I think it's better to just not touch that...

bitonality:
> Turns out the `build.cache` policy really messes up proxy gen so I think it's better to just not touch that...

I'm open to investigating that once we've made some more headway in linux-land. This first pass dependency cache is already a big step in the right direction.

bitonality (approved):

localcc (approved):

PR #512, opened 2024-05-17
fix: Removed most lag in the live view tab, including when you search

UE4SS:
**Description**

We're now rendering only the objects that are visible to the user instead of the entirety of GUObjectArray.
The result is that most, if not all, of the lag is gone.

~~Note 1: This does not apply to search results, with or without `Apply when not searching` enabled.~~
Note 1: This applies to search results as well, but not while `Apply when not searching` is enabled. 
Note 2: Since the lag problem is gone, I also removed the group system.

**Type of change**

**How Has This Been Tested?**

Start the game (preferably a big one), go the live view and scroll down the list, observe the amount of lag.

**Checklist**


Buckminsterfullerene02:
Oh wow that is way better! All seems to work for me, can't find anything not working as expected.

Buckminsterfullerene02 (approved):

UE4SS:
Gonna hold off a minute on this one.
I realized late last night that we might make this work for search results as well.
Will comment when it's ready for merge.

UE4SS:
Alright, this is now ready for review, and another round of testing.
The only thing changed is that the same reduction in lag should now apply to search results if `Apply when not searching` isn't enabled.

Buckminsterfullerene02:
So just to clarify, this PR isn't fixing any of the lag with `Apply when not searching` enabled?

UE4SS:
> So just to clarify, this PR isn't fixing any of the lag with `Apply when not searching` enabled?

Correct.

Buckminsterfullerene02:
Are there any plans to address this lag? Though, the only use case of it, I feel, is when you have a mod that is loading and unloading certain assets so you'll probably be filtered down to those anyway... never had any other reason to use it. Definately wouldn't want to use it when there are thousands of entries in the box. 

Maybe unrelated, but I also get this error in my console when I have it enabled and I have searched for something quite generic ("FSD" in DRG):
`guobjectarray_by_name_iterator: asked to iterate beyond the size of the search result vector (439259 > 15097)`

UE4SS:
> Are there any plans to address this lag? Though, the only use case of it, I feel, is when you have a mod that is loading and unloading certain assets so you'll probably be filtered down to those anyway... never had any other reason to use it. Definately wouldn't want to use it when there are thousands of entries in the box.

Most likely not with this PR.

> Maybe unrelated, but I also get this error in my console when I have it enabled and I search for anything: `guobjectarray_by_name_iterator: asked to iterate beyond the size of the search result vector (439259 > 15097)`

I will look into this, thanks.

Buckminsterfullerene02:
> > Maybe unrelated, but I also get this error in my console when I have it enabled and I search for anything: (something generic): `guobjectarray_by_name_iterator: asked to iterate beyond the size of the search result vector (439259 > 15097)` (something specific with one result): `guobjectarray_by_name_iterator: asked to iterate beyond the size of the search result vector (439259 > 1)`
> 
> @Buckminsterfullerene02 I'm unable to reproduce this. I've tried both in my test game and in DRG and no matter how much I scroll after searching (for `Player`), I never get this message.

You enabled `Apply when not searching` right?

UE4SS:
> > > Maybe unrelated, but I also get this error in my console when I have it enabled and I search for anything: (something generic): `guobjectarray_by_name_iterator: asked to iterate beyond the size of the search result vector (439259 > 15097)` (something specific with one result): `guobjectarray_by_name_iterator: asked to iterate beyond the size of the search result vector (439259 > 1)`
> > 
> > 
> > @Buckminsterfullerene02 I'm unable to reproduce this. I've tried both in my test game and in DRG and no matter how much I scroll after searching (for `Player`), I never get this message.
> 
> You enabled `Apply when not searching` right?

No, I didn't see this mentioned as something I needed to do.

Buckminsterfullerene02:
Apologies, "when I have it enabled" isn't very clear.

UE4SS:
> Apologies, "when I have it enabled" isn't very clear.

It's perfectly clear, but evidently I'm blind.
I should stop reading so quickly.
Anyways, I've found the problem and a fix should be ready soon.

UE4SS:
@Buckminsterfullerene02 The problem should now be fixed.
The problem was that it was possible to enable `Apply when not searching` while you were actually searching.
This meant it was using the wrong iterator.
I've made sure it's not possible to do both these things at the same time, and I also added some tooltips when the user puts themselves in a position where they might try to do that.

Buckminsterfullerene02:
Hmm, I'm not sure I fully understand the point of this option if you can't enable it while having something in the search bar.

UE4SS:
> Hmm, I'm not sure I fully understand the point of this option if you can't enable it while having something in the search bar.

The point of the option is to apply the filters while you're not searching.
For example, if you wanted to get a list of all objects with a property of a certain type.

Buckminsterfullerene02:
Ohhh I get it now, thanks. Can you rename it to `Apply filters when not searching` to make it a bit clearer? Or add a hover tooltip text on it to explain because I straight up never realised that's what it actually does this whole time. I even put this incorrect description in the live view docs... can you also change that 😂 
> Changes the search results automatically, otherwise, you need to press refresh settings.

UE4SS:
> Ohhh I get it now, thanks. Can you rename it to `Apply filters when not searching` to make it a bit clearer? Or add a hover tooltip text on it to explain because I straight up never realised that's what it actually does this whole time. I even put this incorrect description in the live view docs... can you also change that 😂
> 
> > Changes the search results automatically, otherwise, you need to press refresh settings.

Done.

Buckminsterfullerene02 (approved):

PR #517, opened 2024-05-18
build: change third party into a repo system

localcc:
**Description**

This uses the xmake repository system which is the correct way to have packages inside of a project,
allowing us to depend on those packages in other packages as well as improving package xmake.lua
change detection.

**Type of change**

- [x] Other... Please describe: Third-party structure change

**How Has This Been Tested?**

Building the project with new changes after cleaning everything including xrepo using `xrepo remove --all -y`

**Additional context**

Some packages that write their own `xmake.lua` were not able to depend on other packages we've defined ourselves because of it not being a repository, this PR fixes that.


deps/xmake.lua
```diff
@@ -2,3 +2,8 @@
 includes("first")
 -- Third party dependencies
 includes("third")
+add_repositories("third deps/third-repo", { rootdir = get_config("ue4ssRoot") })
```
bitonality, line 5: Should we consider a more descriptive alias for the repo? I'm fine with `third` if we add an inline comment describing what `add_repositories` is doing. 
image

deps/third/xmake.lua
```diff
@@ -1,11 +1 @@
...
-includes("glaze")
-includes("imgui")
-includes("glad")
-includes("raw_pdb")
-
-add_requires("zycore v1.5.0", { debug = is_mode_debug(), configs = {runtimes = get_mode_runtimes()} })
-add_requires("zydis v4.1.0", { debug = is_mode_debug(), configs = {runtimes = get_mode_runtimes()} })
-add_requires("polyhook_2", { debug = is_mode_debug(), configs = {runtimes = get_mode_runtimes()} })
\ No newline at end of file
+includes("glad")
```
bitonality, line 1: Is there an argument to package-ize all our third party deps? Glad has a package def that we can rip off (or install from xrepo)

https://github.com/xmake-io/xmake-repo/blob/0cc9a8d6ce18f77e92b2155f9f87d1509c5b530b/packages/g/glad/xmake.lua#L4

Does our glad do anything special?

localcc: I prefer to keep everything a target when possible, glad is also something that can change based on the configuration we generate it in.

deps/xmake.lua
localcc: I'll just add an inline comment, will also rename the repo to `third-party`

localcc:
as usual this blew up on linux because of casing

bitonality (approved):

narknon (approved):

Issue #526, opened 2024-05-24
[Improvement] Allow vsproj generation of a subset of allowed xmake modes


bitonality:
Mod authors often only want to 
`xmake project -k vsxmake2022 -m Game__Shipping__Win64, Game__Debug__Win64`
I'll see if that's supportable.

Buckminsterfullerene02:
Can this be closed as this feature exists in main?

UE4SS:
> Can this be closed as this feature exists in main?

Yeah I don't see why not.

PR #530, opened 2024-05-27
Advanced mod guide

### LINKED ISSUE #525: [Improvement] Increase flexibility of xmake target settings for mod authors.
Streamline and document the process of when mod authors want to deviate from the *default* `ue4ss.mod` options.

Current def:
```lua

-- This rule is meant to be used by mod targets.
-- Logic that should ONLY apply to mods should be defined in this rule.
-- All other common logic is inherited from the ue4ss.base rule.
-- Example:
-- target("ExampleMod")
--   add_rules("ue4ss.mod")
--   add_includeirs(".")
--   add_files("dllmain.cpp")
rule("ue4ss.mod")
    add_deps("ue4ss.base", {order = true})
    after_load(function(target)
        target:set("kind", "shared")
        target:set("languages", "cxx20")
        target:set("exceptions", "cxx")
        target:add("deps", "UE4SS")
        target:set("group", "mods")
    end)

    on_install(function(target)
        import("mods.install").install(target)
    end)
```

I need to document how mod authors can alter these settings and save their own custom rules if they want. I'm envisioning allowing users to create an extension of `ue4ss.mod` like `ue4ss.mod.cxx_latest` or `ue4ss.mod.custom_runtime` etc. They'll then be able to use their custom rules in any mods in their UE4SS mono-repo.

bitonality:
Clarifies #526 
Fixes #525

### DIFF
--- a/docs/guides/advanced-c++-modding.md
+++ b/docs/guides/advanced-c++-modding.md
@@ -1,0 +1,250 @@
+# Advanced C++ Modding Handbook
+
+## xmake Customization
+
+We provide an `xmake` rule in order to facilitate easy mod setup. The simplest mod `xmake.lua` file looks something like this:
+
+```lua
+ target("SimpleMod")
+   add_rules("ue4ss.mod") -- Adding this rule gives you all the built-in goodies!
+   add_includeirs(".")
+   add_files("dllmain.cpp")
+```
+
+>[!IMPORTANT] What does `ue4ss.mod` get me?
+> Adding the `ue4ss.mod` rule is equivalent to  
+>
+>```lua
+> target("DefaultMod")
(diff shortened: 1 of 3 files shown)


Buckminsterfullerene02:
@bitonality we are using mdbook for docs which means that;
- the new github markdown features like important, tip, caution etc. are not supported by it (as far as I know, I don't recall seeing it in their docs, but they do move fast so maybe they have it now). Otherwise I would've been using them a lot more before...
- when adding new files, you also need to update the page tree in SUMMARY.md. it's pretty self explanatory how that works

For either of the above you can manually merge docs-export and docs folders in the repo and download mdbook.exe to put in there so you can run a local copy to see how your changes look in it.

PR #534, opened 2024-05-29
fix: mappings file dump location

Buckminsterfullerene02:
**Description**

mappings file now dumps to working dir instead of root dir

**Type of change**

**How Has This Been Tested?**

Ran mappings dumper with working dir both in root dir and in root/ue4ss folder

**Checklist**


UE4SS:
Oh uhm... I just realized we didn't mention the backwards compatible nature of the location change in the changelog.
Perhaps we should do that ? That way we don't scare mod manager devs into thinking that they absolutely have to make an update for this.

Buckminsterfullerene02:
Solid shout

UE4SS (approved):

PR #540, opened 2024-05-31
Converts mods.txt to use mods.json

narknon:
Maintains legacy support for mods.txt and favors mods.txt for settings
mods.txt should only be favored for one minor release cycle post 4.0
Our packaging script will need to be updated, but after mods.txt and ue4sssettings are converted to json it should be able to be simplified significantly.

### DIFF
--- a/UE4SS/include/UE4SSProgram.hpp
+++ b/UE4SS/include/UE4SSProgram.hpp
@@ -20,6 +20,8 @@
 #include <SettingsManager.hpp>
 #include <Unreal/Core/Containers/Array.hpp>
 #include <Unreal/UnrealVersion.hpp>
+#include <glaze/glaze.hpp>
+#include <glaze/core/macros.hpp>
 
 // Used to set up ImGui context and allocator in DLL mods
 #define UE4SS_ENABLE_IMGUI()                                                                                                                                   \
@@ -207,8 +209,18 @@ namespace RC
         auto fire_dll_load_for_cpp_mods(std::wstring_view dll_name) -> void;
 
       public:
+        struct ModData {
+            std::string mod_name;
+            bool mod_enabled;
+
+            GLZ_LOCAL_META(ModData, mod_name, mod_enabled);
+        };
+        
         auto init() -> void;
         auto is_program_started() -> bool;
+        static auto read_mods_json(std::string enabled_mods_file, std::vector<ModData>& mod_data_vector) -> void;
+        static auto write_mods_json(std::string enabled_mods_file, std::vector<ModData>& mod_data_vector) -> void;
+        static auto convert_legacy_mods_file(StringType legacy_enabled_mods_file, std::vector<ModData>& mod_data_vector) -> void;
         auto reinstall_mods() -> void;
         auto get_object_dumper_output_directory() -> const File::StringType;
         RC_UE4SS_API auto get_module_directory() -> File::StringViewType;
--- a/UE4SS/src/UE4SSProgram.cpp
+++ b/UE4SS/src/UE4SSProgram.cpp
@@ -1115,67 +1115,104 @@ namespace RC
         }
     }
 
+    auto UE4SSProgram::read_mods_json(std::string enabled_mods_file, std::vector<ModData>& mod_data_vector) -> void
+    {
+        auto ec = glz::read_file_json(mod_data_vector, enabled_mods_file, std::string{});
+    }
+
+    auto UE4SSProgram::write_mods_json(std::string enabled_mods_file, std::vector<ModData>& mod_data_vector) -> void
+    {
+        std::string mod_data_buffer;
+        glz::write<glz::opts{.prettify = true}>(mod_data_vector, mod_data_buffer);
+        glz::error_code ec = glz::buffer_to_file(mod_data_buffer, enabled_mods_file);
+    }
+
+    auto UE4SSProgram::convert_legacy_mods_file(StringType legacy_enabled_mods_file, std::vector<ModData>& mod_data_vector) -> void
+    {
+        Output::send(STR("Converting legacy mods.txt to mods.json...\n"));
+        
+        // 'mods.txt' exists, lets parse it
+        std::wifstream mods_stream{legacy_enabled_mods_file};
+
+        std::wstring current_line;
+        while (std::getline(mods_stream, current_line))
+        {
+            // Don't parse any lines with ';'
+            if (current_line.find(L";") != current_line.npos)
+            {
+                continue;
+            }
+
+            // Don't parse if the line is impossibly short (empty lines for example)
+            if (current_line.size() <= 4)
+            {
+                continue;
+            }
+
+            // Remove all spaces
+            auto end = std::remove(current_line.begin(), current_line.end(), L' ');
+            current_line.erase(end, current_line.end());
+
+            // Parse the line into something that can be converted into proper data
+            std::wstring mod_name = explode_by_occurrence(current_line, L':', 1);
+            std::wstring mod_enabled = explode_by_occurrence(current_line, L':', ExplodeType::FromEnd);
+            bool mod_enabled_out = !mod_enabled.empty() && mod_enabled[0] == L'1';
+
+            auto it = std::find_if(mod_data_vector.begin(), mod_data_vector.end(),
+                           [&mod_name](const ModData& element) {
+                               return element.mod_name == to_string(mod_name);
+                           });
+
+            if (it == mod_data_vector.end()) {
+                mod_data_vector.push_back({to_string(mod_name), mod_enabled_out});
+            }
+        }
+        
+        Output::send(STR("mods.txt entries converted to mods.json.  Please delete mods.txt if no mod managers rely on it to avoid conflicting with changes to the json...\n"));
+    }
+
+
     template <typename ModType>
     auto start_mods() -> std::string
     {
         ProfilerScope();
-        // Part #1: Start all mods that are enabled in mods.txt.
-        Output::send(STR("Starting mods (from mods.txt load order)...\n"));
-
+        // Part #1: Start all mods that are enabled in mods.json.
         std::filesystem::path mods_directory = UE4SSProgram::get_program().get_mods_directory();
-        std::wstring enabled_mods_file{mods_directory / "mods.txt"};
-        if (!std::filesystem::exists(enabled_mods_file))
+        StringType legacy_enabled_mods_file{mods_directory / "mods.txt"};
+        std::filesystem::path enabled_mods_file{mods_directory / "mods.json"};
+        std::vector<UE4SSProgram::ModData> mod_data_vector{};
+
+        if (std::filesystem::exists(legacy_enabled_mods_file))
         {
-            Output::send(STR("No mods.txt file found...\n"));
+            UE4SSProgram::convert_legacy_mods_file(legacy_enabled_mods_file, mod_data_vector);
         }
-        else
+        if (std::filesystem::exists(enabled_mods_file))
         {
-            // 'mods.txt' exists, lets parse it
-            std::wifstream mods_stream{enabled_mods_file};
+            UE4SSProgram::read_mods_json(enabled_mods_file.string(), mod_data_vector);
+        }
+        
+        UE4SSProgram::write_mods_json(enabled_mods_file.string(), mod_data_vector);
 
-            std::wstring current_line;
-            while (std::getline(mods_stream, current_line))
+        Output::send(STR("Starting mods (from mods.json load order)...\n"));
+        for (auto it = mod_data_vector.begin(); it != mod_data_vector.end(); ++it)
+        {
+            auto mod = UE4SSProgram::find_mod_by_name<ModType>(it->mod_name, UE4SSProgram::IsInstalled::Yes);
+            if (!mod || !dynamic_cast<ModType*>(mod))
             {
-                // Don't parse any lines with ';'
-                if (current_line.find(L";") != current_line.npos)
-                {
-                    continue;
-                }
+                continue;
+            }
 
-                // Don't parse if the line is impossibly short (empty lines for example)
-                if (current_line.size() <= 4)
-                {
-                    continue;
-                }
-
-                // Remove all spaces
-                auto end = std::remove(current_line.begin(), current_line.end(), L' ');
-                current_line.erase(end, current_line.end());
-
-                // Parse the line into something that can be converted into proper data
-                std::wstring mod_name = explode_by_occurrence(current_line, L':', 1);
-                std::wstring mod_enabled = explode_by_occurrence(current_line, L':', ExplodeType::FromEnd);
-
-                auto mod = UE4SSProgram::find_mod_by_name<ModType>(mod_name, UE4SSProgram::IsInstalled::Yes);
-                if (!mod || !dynamic_cast<ModType*>(mod))
-                {
-                    continue;
-                }
-
-                if (!mod_enabled.empty() && mod_enabled[0] == L'1')
-                {
-                    Output::send(STR("Starting {} mod '{}'\n"), std::is_same_v<ModType, LuaMod> ? STR("Lua") : STR("C++"), mod->get_name().data());
-                    mod->start_mod();
-                }
-                else
-                {
-                    Output::send(STR("Mod '{}' disabled in mods.txt.\n"), mod_name);
-                }
+            if (it->mod_enabled)
+            {
+                Output::send(STR("Starting {} mod '{}'\n"), std::is_same_v<ModType, LuaMod> ? STR("Lua") : STR("C++"), mod->get_name().data());
+                mod->start_mod();
+            }
+            else
+            {
+                Output::send(STR("Mod '{}' disabled in mods.json.\n"), to_wstring(it->mod_name));
             }
         }
-
-        // Part #2: Start all mods that have enabled.txt present in the mod directory.
-        Output::send(STR("Starting mods (from enabled.txt, no defined load order)...\n"));
+        
 
         for (const auto& mod_directory : std::filesystem::directory_iterator(mods_directory))
         {
--- a/assets/Mods/mods.json
+++ b/assets/Mods/mods.json
@@ -1,0 +1,42 @@
+[
+   {
+      "mod_name": "CheatManagerEnablerMod",
+      "mod_enabled": true
+   },
+   {
+      "mod_name": "ActorDumperMod",
+      "mod_enabled": false
+   },
+   {
+      "mod_name": "ConsoleCommandsMod",
+      "mod_enabled": true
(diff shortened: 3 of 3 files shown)


UE4SS:
I can see why mods.txt would be better as a standardized format, but I don't see why UE4SS-settings needs to change from ini to json.
If there's no actual benefit here beyond "all our stuff is json", then we're just adding potential difficulties for the end user.
I know this PR doesn't modify UE4SS-settings, but when/if this happens it would be useful if whoever does it goes into detail about why the change is needed.

UE4SS:
> > I can see why mods.txt would be better as a standardized format, but I don't see why UE4SS-settings needs to change from ini to json. If there's no actual benefit here beyond "all our stuff is json", then we're just adding potential difficulties for the end user. I know this PR doesn't modify UE4SS-settings, but when/if this happens it would be useful if whoever does it goes into detail about why the change is needed.
> 
> JSON schema allows for representing objects and arrays in the file (something ini files struggle with). If the current settings are only kvps then I see your point.
> 
> Have we considered if we _should_ be making the settings file more than key value pairs? Some of the beauty of the ini format (as a settings mechanism) is that it's pretty obvious when it's being misused for things it shouldn't be.
> 
> Another thing to consider is that we lose the ability to ship comments/descriptions about settings within the settings file if we switch to json. Ini, toml, yaml, all have the ability to comment within the file but that's one of the key "weaknesses" of json as a configuration schema. (It's a strength for serialization/requests depending on who you ask since it makes the schema easier to regulate when using json as request payloads/http responses over the wire...)
> 
> TOML and yaml are both alternatives that support comments but have their own histories/quirks.
> 
> Just making sure that we're thinking about what the right tool for the right job is.

We can "fake" comments in JSON by having them as a field so that shouldn't be a huge problem, and I do think comments are crucial for the settings file.
I like ini because it's very simple and easy to read.
With JSON, you get more advanced features, but it's harder to read, there's substantial visual cluttering.
While we do have some basic support for arrays (section as ordered list & section entries without values) in our ini system, that's not part of any kind of ini standard so it would potentially make it harder for anyone to programmatically parse the file, though I don't believe we use any of those features in UE4SS-settings.ini.

My opinion is that we should stay away from JSON or any other format for the settings file unless we desperately need more advanced features, and we should also take into account whether the setting we're implementing that requires more advanced features actually belongs in the settings file.

Example of how a basic ini structure would look like, and the equivalent in JSON:
```ini
[Section1]
; Description of A
A = 1
; Description of B
B = 2
; Description of C
C = 3

[Section2]
; Description of A
A = 1
; Description of B
B = 2
; Description of C
C = 3
```

```json
{
    "Section1": {
        "Description of A": null,
        "A": 1,
        "Description of B": null,
        "B": 2,
        "Descriptio of C": null,
        "C": 3
    },
    "Section2": {
        "Description of A": null,
        "A": 1,
        "Description of B": null,
        "B": 2,
        "Descriptio of C": null,
        "C": 3
    }
}
```

bitonality:
I'd personally avoid doing the json comment by using the adding a new key hack. Whenever I see that mechanism being used, it screams "I'm misusing the intent of this schema". Json was designed as and still is a schema for data-interchange.

bitonality:
> > I'd personally avoid doing the json comment by using the adding a new key hack. Whenever I see that mechanism being used, it screams "I'm misusing the intent of this schema". Json was designed as and still is a schema for data-interchange.
> 
> Wouldn't that interfere with programs parsing the file with the assumption that it's standard json ? That's a concern because people do actually parse the UE4SS-settings file.

Let me clarify - i strongly advise against hacking in comment functionality into json through any means (new kvp, preprocessor, etc)

UE4SS:
> > > I'd personally avoid doing the json comment by using the adding a new key hack. Whenever I see that mechanism being used, it screams "I'm misusing the intent of this schema". Json was designed as and still is a schema for data-interchange.
> > 
> > 
> > Wouldn't that interfere with programs parsing the file with the assumption that it's standard json ? That's a concern because people do actually parse the UE4SS-settings file.
> 
> Let me clarify - i strongly advise against hacking in comment functionality into json through any means (new kvp, preprocessor, etc)

Oh okay, I must've misunderstood, good to have clarification.
I generally agree.

bitonality:
# Schema Comparison

**Don't use this chart as the source of truth in terms of deciding what schema to use.** I just wanted to provide some data points about file size bloat for each schema type in the context of `mods.txt`. I want to ensure that we're looking to the future and taking scale into account so we can make informed decisions about the strengths/limitations of whatever schema we end up using.

# Json

```json
[
   {
      "mod_name": "CheatManagerEnablerMod",
      "mod_enabled": true
   },
   {
      "mod_name": "ActorDumperMod",
      "mod_enabled": false
   },
]
```

# Toml

```toml
"CheatManagerEnablerMod" = true
"ActorDumperMod" = false
```

# Yaml

```yaml
CheatManagerEnablerMod: yes
ActorDumperMod: no
```

# Xml

```xml
<mod name="CheatManagerEnablerMod" isEnabled="true"/>
<mod name="ActorDumperMod" isEnabled="false"/>
```

# Mods.txt

```txt
CheatManagerEnablerMod : 1
ActorDumperMod : 0
```

# Napkin Math

| Format | Up-front bytes | Per-Entry Bytes | Comment Bytes |
|---|---|---|---|
| JSON | 4 | 59-60 + N | 0 |
| TOML | 0 | 10-11 + N | 1+C |
| YML | 0 | 6 + N | 1+C|
| XML | 0 | 32-33 + N |  7 + C|
| TXT | 0 | 5 + N | 3 + C|

# Graph
SchemaComparison

narknon:
I don't think we have to worry about file bloat for mods.txt/json.  It should only ever be for enable/disable and any additional metadata would go elsewhere.  Toml may be "ideal" for very simple files and yaml for complicated ones, but that is two separate libs that we'd need to pull in and that anyone interacting with our files would need to pull in, when json is a good middle ground for both. 

Wrong place for general json settings discussion, but we iterated over several methods of setting up the settings file in json a long time ago, and made helpers to make commenting easy. 

Our comment templates: 
image 

There is also jsonc:  
image 

JSON is originally intended to be a data interchange format, sure, but the original intention is not relevant.  In my opinion (and many others' opinions) it is very readable as a config format, and our library has powerful syntax error logging.

narknon:
Also of note for the settings file is that we will likely have a GUI for the settings at some point, but even if we do not have a separate program, the debugging GUI will have a tab.  The only reason it doesn't already is because the current format would make it much more annoying to write changes out.

UE4SS:
> Our comment templates: image

Does this support multi-line comments ?
We currently have multi-line comments in the ini file and horizontal scrolling is terrible so I think multi-line comments is something we should support if possible.

Does it support a field for the default value ? Could be part of a multi-line comment.
The default value provides an easy way for the user to experiment with values without having to pull up the file on github whenever they want to revert to the default.
I consider it valuable information.

narknon:
I believe CC made functionality for multiline, yes.  This is another thing that got stuck in decision making limbo due to arguments, so it was long enough ago that I don't fully remember.

narknon:
Yes, there was support for multiline. https://github.com/UE4SS-RE/RE-UE4SS/blob/main/deps/first/Constructs/include/Constructs/Annotated.hpp

UE4SS:
I think runtime modifying of the settings is a good enough reason to change from our custom ini parser since that doesn't support modifying the file, but that brings into question backwards compatibility.
We could implement it similarly to the file location changes, but we'd have to disable runtime modifying if the ini file is being used instead of the new one.
But regardless, this isn't covered in this PR so unless you intend to expand it, I think we can move on from this discussion, and bring it back up when someone wants to implement changes to the settings file, or alternatively move this to an issue or discussion (with reference to here) if you think more discussion is needed.

narknon:
Last time I implemented this I had it read out the old file and convert it to json, similar to what I do in this PR. I believe the in progress branch still exists as JSONSettings, but I can't recall what the context of the last push to it was.

But yeah, this can be made into an issue or separate discussion.

UE4SS:
> Last time I implemented this I had it read out the old file and convert it to json, similar to what I do in this PR. I believe the in progress branch still exists as JSONSettings, but I can't recall what the context of the last push to it was.

The problem with converting it is that it doesn't address the problem of third-party parsers expecting an ini file with the ini format.

narknon:
I'm 99% sure there are no third party parsers for it.

UE4SS:
> I'm 99% sure there are no third party parsers for it.

We can't know for sure if there are, but it's generally a bad idea to break compatibility, and there is a better solution so we should use it.

narknon:
I haven't even found a third party mod manager that actually uses mods.txt rather than enabled.txt, let alone the settings.  I'd literally eat my shoe if there is one.  But this PR favors the old mods.txt when updating the new one, though I am considering changing that now that I have done more research into whether any mod managers use mods.txt.  A similar thing could happen here.  Best practices shouldn't be used to the point of creating problems where there are none.

bitonality:
> I don't think we have to worry about file bloat for mods.txt/json.

I agree that it's negligible in this case. I just wanted to make sure that we understand that we're increasing the size of that file by ~10x and that we're okay with that (which it looks like we are unless anyone wants to object).

UE4SS:
> > I don't think we have to worry about file bloat for mods.txt/json.
> 
> I agree that it's negligible in this case. I just wanted to make sure that we understand that we're increasing the size of that file by ~10x and that we're okay with that (which it looks like we are unless anyone wants to object).

I mean worst case (and unlikely), someone will run a couple hundred mods.
How long would it take for glaze to parse a file that big ? Surely not long enough to be a problem ? It wouldn't be that big of a file.
Is that the main concern ? If so, it's not a big problem even if it does end up taking a huge amount of time like a couple of seconds extra to load the game, as long as it doesn't go any further than that.

bitonality:
> > > I don't think we have to worry about file bloat for mods.txt/json.
> > 
> > 
> > I agree that it's negligible in this case. I just wanted to make sure that we understand that we're increasing the size of that file by ~10x and that we're okay with that (which it looks like we are unless anyone wants to object).
> 
> I mean worst case (and unlikely), someone will run a couple hundred mods. How long would it take for glaze to parse a file that big ? Surely not long enough to be a problem ? It wouldn't be that big of a file. Is that the main concern ? If so, it's not a big problem even if it does end up taking a huge amount of time like a couple of seconds extra to load the game, as long as it doesn't go any further than that.

I agree that it won't be a problem for a sane amount of mods. I just wanted to make sure we are planning on getting benefits from switching to offset the increase in file size. From what I've read on this discussion: readability, parsing libs, other mentioned benefits all outweigh the negligible size increase.

I just wanted to make sure we are getting tangible benefits and not just increasing the file size without any upsides in return. Apologies if it came off as being pedantic about size alone.

narknon:
> 
> There will be confusion from users that end up having both mods.txt and mods.json. They won't know which one to edit or the repercussions of modifying the json file. The json file will be overwritten with the contents of the txt file so any changes to the json file will be discarded, and this is harmful to the user.
> 
> It is backwards compatible assuming that it always converts and thus fully discards the current json to avoid appending the data multiple times in it, but it has the problem mentioned above.
> 
> EDIT: This is all assuming that glaze isn't smart about it and just overwrites the entire file instead of setting each value one by one leaving existing (and unknown) values unchanged.


I suppose we could do favor ini/txt if not default settings, if ini default but json isn't favor json for each setting.. At this point I'm giving up on moving the convo for the ue4ss-settings from this PR since we're this deep into it. 


In my opinion, the full backwards compat favoring settings.ini/mods.txt should last at most one minor release cycle post 4.0, or a couple months, whichever is longer.  After that, the converter should be left in until 5.0, but the legacy files shouldn't be read if the json exists.  It's a bit of a bandaid rip but with our release cycle I think that gives ample time for any mod managers that would ever make the switch to make the switch. 

Glz will always read/write from the struct, so it's up to us how we want to populate the data in the struct from the existing ini/txt or jsons and which gets favored.

UE4SS:
Can you rename `legacy_enabled_mods_file`, `enabled_mods_file`, and any other named references to mods.txt/json to something that doesn't clash with our concept of `enabled.txt` ?
I got confused by this naming at first when looking at your code.

narknon:
> Can you rename `legacy_enabled_mods_file`, `enabled_mods_file`, and any other named references to mods.txt/json to something that doesn't clash with our concept of `enabled.txt` ? I got confused by this naming at first when looking at your code.

Sure though enabled_mods_file was the original name of mods.txt, not something I changed.

UE4SS:
> > Can you rename `legacy_enabled_mods_file`, `enabled_mods_file`, and any other named references to mods.txt/json to something that doesn't clash with our concept of `enabled.txt` ? I got confused by this naming at first when looking at your code.
> 
> Sure though enabled_mods_file was the original name of mods.txt, not something I changed.

Understood.
This was introduced before enabled.txt so that's why it was never changed but since we're making changes to the code now, we might as well make it less confusing.

UE4SS/src/UE4SSProgram.cpp
```diff
@@ -1117,67 +1117,127 @@ namespace RC
...
+            wchar_t* converted_method_name = new wchar_t[count];
+
+            size_t num_of_char_converted = 0;
+            mbstowcs_s(&num_of_char_converted, converted_method_name, count, descriptive_error.data(), count);
+
+            auto converted = File::StringViewType(converted_method_name);
+
+            delete[] converted_method_name;
+                
+            Output::send<LogLevel::Error>(STR("{}\n\nError parsing mods.json file, please fix the file...\n"), converted);
```
UE4SS, line 1138: What is this ?
Why aren't we using `to_wstring` ?

narknon:
Before I push it, does mod_list_file/legacy work?

UE4SS:
> Before I push it, does mod_list_file/legacy work?

Yes, that sounds fine.

UE4SS/src/UE4SSProgram.cpp
```diff
@@ -1117,67 +1117,127 @@ namespace RC
...
+    {
+        std::string mod_data_buffer;
+        glz::write<glz::opts{.prettify = true}>(mod_data_vector, mod_data_buffer);
+        glz::error_code ec = glz::buffer_to_file(mod_data_buffer, enabled_mods_file);
+    }
+    
+    auto UE4SSProgram::convert_legacy_mods_file(StringType legacy_enabled_mods_file, std::vector<ModData>& mod_data_vector) -> void
+    {
+        Output::send(STR("Converting legacy mods.txt to mods.json...\n"));
+        std::wifstream mods_stream = File::open_file_skip_BOM(legacy_enabled_mods_file);
```
UE4SS: Should use `StreamType` here instead of `std::wifstream`.

UE4SS/include/UE4SSProgram.hpp
```diff
@@ -207,8 +209,20 @@ namespace RC
         auto fire_dll_load_for_cpp_mods(std::wstring_view dll_name) -> void;
 
       public:
+        struct ModData {
+            std::string mod_name;
+            bool mod_enabled;
+
+            GLZ_LOCAL_META(ModData, mod_name, mod_enabled);
+        };
+        std::vector<ModData> global_mod_data_vector;
```
UE4SS, line 218: Unused variable ?
I can't find any usages in this PR.

UE4SS/src/UE4SSProgram.cpp
```diff
@@ -1117,67 +1117,127 @@ namespace RC
...
+    auto UE4SSProgram::convert_legacy_mods_file(StringType legacy_mod_list_file, std::vector<ModData>& mod_data_vector) -> void
+    {
+        Output::send(STR("Converting legacy mods.txt to mods.json...\n"));
+        std::wifstream mods_stream = File::open_file_skip_BOM(legacy_mod_list_file);
+        try {
+            mods_stream = File::open_file_skip_BOM(legacy_mod_list_file);
+        } catch (const std::exception& e) {
+            std::cerr << e.what() << std::endl;
+        }
+        std::wstring current_line;
```
UE4SS, line 1158: Use `StringType`.

UE4SS/src/UE4SSProgram.cpp
```diff
@@ -1117,67 +1117,127 @@ namespace RC
...
+    
+    auto UE4SSProgram::convert_legacy_mods_file(StringType legacy_mod_list_file, std::vector<ModData>& mod_data_vector) -> void
+    {
+        Output::send(STR("Converting legacy mods.txt to mods.json...\n"));
+        std::wifstream mods_stream = File::open_file_skip_BOM(legacy_mod_list_file);
+        try {
+            mods_stream = File::open_file_skip_BOM(legacy_mod_list_file);
+        } catch (const std::exception& e) {
+            std::cerr << e.what() << std::endl;
+        }
```
UE4SS, line 1157: This is bad, don't throw here, just let the exception fall through to main like we do with all fatal exceptions.
The way you've currently set this up will cause the error to not actually be logged, it will just be ignored and probably crash as you're still continuing to operate on the stream even though there was an error with it causing the stream to be invalid.

UE4SS/src/UE4SSProgram.cpp
```diff
@@ -1117,67 +1117,127 @@ namespace RC
...
+            // Remove all spaces
+            auto end = std::remove(current_line.begin(), current_line.end(), L' ');
+            current_line.erase(end, current_line.end());
 
-                // Parse the line into something that can be converted into proper data
-                std::wstring mod_name = explode_by_occurrence(current_line, L':', 1);
-                std::wstring mod_enabled = explode_by_occurrence(current_line, L':', ExplodeType::FromEnd);
+            // Parse the line into something that can be converted into proper data
+            std::wstring mod_name = explode_by_occurrence(current_line, L':', 1);
+            std::wstring mod_enabled = explode_by_occurrence(current_line, L':', ExplodeType::FromEnd);
```
UE4SS, line 1180: Use `StringType`, and `STR('...')` instead of `L'...'`.

UE4SS/src/UE4SSProgram.cpp
```diff
@@ -1117,67 +1117,127 @@ namespace RC
...
+        if (std::filesystem::exists(legacy_mod_list_file))
+        {
+            UE4SSProgram::convert_legacy_mods_file(legacy_mod_list_file, mod_data_vector);
+        }
+        
+        UE4SSProgram::write_mods_json(mod_list_file.string(), mod_data_vector);
+
+        
+        Output::send(STR("Starting mods (from mods.json load order)...\n"));
+        for (auto it = mod_data_vector.begin(); it != mod_data_vector.end(); ++it)
```
UE4SS, line 1222: Any reason why we're using the more complicated iterator method of looping here instead of the simpler
`for (const auto& mod_data : mod_data_vector)` ?

UE4SS (changes requested):
Please run clang format.
Note that I haven't yet tested this PR.

PR #542, opened 2024-05-31
Rewrite build.py -> release.py and fix bugs

### LINKED ISSUE #538: Mods folder
In latest experimental packaging, Mods folder is copied in as a subfolder of itself.  I believe this is ever since the pathing changes.

### LINKED ISSUE #539: Settings
Since the path changes PR merge, the settings in dev releases are the same as release versions.

Buckminsterfullerene02:
**Description**

Fixes the outstanding bugs currently in main. Rewritten for readability. Some extra error handling, adds concurrency groups to actions for more robustness, as currently the action fails in a potentially bad way when multiple are executed in a row (such as multiple merges happening in a flurry). 

Fixes #538 
Fixes #539 

**Type of change**

**How Has This Been Tested?**

Local testing:
- Zip files match as expected
- Tested case when kismetdebugger compiled
- Tested case when kismetdebugger not compiled

Release:
- Zip files match as expected

**Checklist**


narknon:
Release version mods.txt including the cpp mods that aren't included in release is intended, in case release and dev zips get mixed together without overwriting mods.txt

bitonality (commented):
Looks good for restoring the correct behavior. I notated my opinions about the fragility of this current script, but I figure it's best to restore release functionality before we address the re-tooling of this script.

https://github.com/UE4SS-RE/RE-UE4SS/commit/162b81c2aaae2af316e56ac186feaa3e44e258de

bitonality:
> Release version mods.txt including the cpp mods that aren't included in release is intended, in case release and dev zips get mixed together without overwriting mods.txt

Was not aware that this was the intended behavior - redacting my previous statement that indicated the correct behavior is restored.

bitonality:
https://github.com/UE4SS-RE/RE-UE4SS/issues/543

Adding this for tracking purposes, but the completion of this PR should not qualify as the full fledged fix for #543

Buckminsterfullerene02:
`release.py`'s git history is restored now

.github/workflows/experimental.yml
bitonality, line 1: We really ought to lock down the python version for this workflow instead of `3.x`ing

.github/workflows/experimental.yml
```diff
@@ -57,9 +57,15 @@ jobs:
           xmake build
 
       - name: Package
-        run: python tools/buildscripts/build.py package -e
+        id: package
+        run: python tools/buildscripts/release.py package -e
+
+      - name: Check for Package Step Failure
```
bitonality, line 63: The default behavior for github workflows is to give up and prevent further execution if any step fails. Am I missing the point of the manual checks?

.github/workflows/release.yml
```diff
@@ -51,16 +51,23 @@ jobs:
           xmake build
 
       - name: Package
-        run: python tools/buildscripts/build.py package
+        id: package
+        run: python tools/buildscripts/release.py package
+
+      - name: Check for Package Step Failure
+        if: ${{ steps.package.outcome == 'failure' }}
+        run: exit 1
```
bitonality, line 59: Likewise with these checks, am I missing the point?

tools/buildscripts/release.py
```diff
@@ -0,0 +1,338 @@
...
+    "Dev" refers to the zip that is used for development. It should be in the format zDEV-UE4SS_<version>.zip and contains all the files, pdbs and docs.
+    """
+
+    def __init__(self, args):
+        self.args = args
+        self.release_output = 'release'
+        self.setup()
+
+    def setup(self):
+        shutil.rmtree(self.release_output, ignore_errors=True)
```
bitonality, line 204: I'm not sure I agree with ignoring errors. Since this is packaging stuff that is intended for distribution, I think it's better to err on the side of caution when it comes to optional error ignoring. Is it okay to proceed with packaging the project if we don't have a clean release output directory?

tools/buildscripts/release.py
```diff
@@ -0,0 +1,338 @@
...
+    def copy_docs(self):
+        if self.is_dev_release and os.path.exists('docs'):
+            shutil.copytree('docs', os.path.join(self.ue4ss_dir, 'Docs'))
+
+    def modify_changelog(self):
+        changelog_path = self.ue4ss_dir + '/Changelog.md'
+        if os.path.exists(changelog_path): ready_changelog_for_release(changelog_path)
+
+    def package_release(self):
+        version = subprocess.check_output(['git', 'describe', '--tags']).decode('utf-8').strip()
```
bitonality, line 183: Nitpick: It might be worth catching the error here to make it a little easier for people investigating failed releases? The current output (when erroring) on my machine is 

```
fatal: No names found, cannot describe anything.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python39\lib\subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "C:\Python39\lib\subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', 'describe', '--tags']' returned non-zero exit status 128.
```

Pruning and shallow cloning in the checkout step would potentially cause the git describe --tags to not work as expected and so we probably should handle it gracefully downstream of that step?

tools/buildscripts/release.py
```diff
@@ -0,0 +1,338 @@
...
+        shutil.copy(dwmapi_dll_path, self.staging_dir)
+        shutil.copy(ue4ss_dll_path, self.ue4ss_dir)
+        if self.is_dev_release: shutil.copy(ue4ss_pdb_path, self.ue4ss_dir)
+
+    def copy_docs(self):
+        if self.is_dev_release and os.path.exists('docs'):
+            shutil.copytree('docs', os.path.join(self.ue4ss_dir, 'Docs'))
+
+    def modify_changelog(self):
+        changelog_path = self.ue4ss_dir + '/Changelog.md'
```
bitonality, line 179: Use os.path.join instead of string appending

tools/buildscripts/release.py
```diff
@@ -0,0 +1,338 @@
...
+        if not self.is_dev_release:
+            for key, value in mods_to_disable_in_release.items():
+                pattern = rf'(^{key}\s*:).*?$'
+                content = re.sub(pattern, rf'\1 {value}', content, flags=re.MULTILINE)
+
+        with open(mods_path, mode='w', encoding='utf-8-sig') as file:
+            file.write(content)
+
+    def modify_mods_json(self):
+        mods_path = os.path.join(self.ue4ss_dir, 'Mods/mods.json')
```
bitonality, line 150: `os.path.join(self.ue4ss_dir, 'Mods', 'mods.json')`

bitonality (changes requested):

.github/workflows/release.yml
bitonality, line 1: This and experimental NEED to leverage the github workflow concurrency setting to protect against simultaneous runs trying to execute.

bitonality:
Need to see the github concurrency setting added to the workflow files + python version locked down before I personally approve. Rest of the comments are nice if you get to them, but not necessarily worth blocking for.

.github/workflows/experimental.yml
Buckminsterfullerene02: Just cos I thought that actions didn't automatically fail if the script does, but I was misinformed whenever that was (oops, I have been doing this for a while)

bitonality (approved):
Thanks for the changes.

Buckminsterfullerene02:
@narknon do you want to do final check of the output files? Release is `-124-ge0fe418`

narknon:
Output looks good.  Do we still have the bug with the first line of mods.txt not activating mods? Maybe it was fixed with the json switch?

Buckminsterfullerene02:
> Output looks good. Do we still have the bug with the first line of mods.txt not activating mods? Maybe it was fixed with the json switch?

I don't know but that's out of scope of this PR anyway.

bitonality (approved):

PR #551, opened 2024-06-02
fix: crash dumps are sent to working directory

Buckminsterfullerene02:
**Description**

Fixes yet another missed case with #506 

**Type of change**

**How Has This Been Tested?**

It hasn't, I need someone to test it by forcecrashing the game to make it generate a dump report. Idk how to crash it to generate dmp report, I thought maybe something like dereferencing a null pointer might work but that just crashes with no report.

### DIFF
--- a/UE4SS/src/CrashDumper.cpp
+++ b/UE4SS/src/CrashDumper.cpp
@@ -3,8 +3,9 @@
 #include <chrono>
 #include <string>
 #include <format>
-#include <filesystem>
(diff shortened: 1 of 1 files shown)


UE4SS:
I was able to generate a dmp file by changing the engine version to the wrong one.
The file was generated in the `<exe-dir>/ue4ss` directory.

UE4SS (approved):

PR #558, opened 2024-06-11
Port/inputsource: Refactor Abstracted Input Handler

Yangff:
**Description**

The original Input Handler only accepted input from Win32. In the refactoring we have defined the input source as Input Source, which allows different input sources to be defined for different scenarios. QueueInputSource allows messages from other threads to generate keyboard input and be dispatched and responded to by the `process_event` thread via the SPSC queue.

Almost all of the definitions remain unchanged, with the only exception that `get_events()` has been replaced by `get_events_safe()`, which allows the use of mutex protection when accessing `m_key_set`.

ModifierKeys are re-implemented using bitwise operations on `uint32_t`, which simplifies storing and comparing modifier keys.

This PR depends on the string change so you can see those commits...

**Type of change**

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

**How Has This Been Tested?**

Compiled and tested on Palworld, may need extra tests to ensure keys are responsed correctly as I only tested basic ones.

**Checklist**

**Additional context**


UE4SS:
This PR is effectively impossible to review in its current state because it's based on #557.
Looking past the #557 changes to try to find what this PR actually changes is not reasonable.
My suggestion is to leave it alone for now, and when #557 has been merged, we can rebase this one back onto main to make the review process more sane.

UE4SS:
Superseded by #749

PR #559, opened 2024-06-11
Port/bugfix 1

Yangff:
**Description**

~~Sorry about this merge order, it's a PR that has almost nothing to do with the previous PR, it just fixes two bugs in the code. you can cherrypick them. I didn't do it because it would affect the history and stash pop of the code I'm currently working on and cause some weird things. Please just look at the last two commits.~~

m_shared_functions and m_mods should be member variables (as annotataed `m_`)

~~File opened with `m_always_create_file` should use `Writing` instead of `Appending`, as stated in `NewFileDevice.hpp`~~
> ~~Identical to FileDevice except it deletes the file & re-creates it before outputting anything~~

**Type of change**

**How Has This Been Tested?**

Compiled

**Checklist**

I don't think this need any document?
As for dependent changes, you can choose if to cherry pick this or merge all my port/* in order. I'm making sure each of my PR compile and runable so they are based on previous PRs.. 

**Additional context**

### DIFF
--- a/UE4SS/include/UE4SSProgram.hpp
+++ b/UE4SS/include/UE4SSProgram.hpp
@@ -144,9 +144,9 @@ namespace RC
         uint64_t m_hook_trampoline_load_library_ex_w;
 
       public:
-        static inline std::vector<std::unique_ptr<Mod>> m_mods;
+        std::vector<std::unique_ptr<Mod>> m_mods;
 
-        static inline RecognizableStruct m_shared_functions{};
+        RecognizableStruct m_shared_functions{};
 
         static inline UE4SSProgram* s_program{};
 
--- a/UE4SS/src/Mod/LuaMod.cpp
+++ b/UE4SS/src/Mod/LuaMod.cpp
@@ -3314,7 +3314,7 @@ Overloads:
             return;
         }
 
-        for (const auto& mod : UE4SSProgram::m_mods)
+        for (const auto& mod : UE4SSProgram::get_program().m_mods)
         {
             if (auto cpp_mod = dynamic_cast<CppMod*>(mod.get()); cpp_mod && mod->is_started())
             {
@@ -3334,7 +3334,7 @@ Overloads:
             return;
         }
 
-        for (const auto& mod : UE4SSProgram::m_mods)
+        for (const auto& mod : UE4SSProgram::get_program().m_mods)
         {
             if (auto cpp_mod = dynamic_cast<CppMod*>(mod.get()); cpp_mod && mod->is_started())
             {
--- a/UE4SS/src/UE4SSProgram.cpp
+++ b/UE4SS/src/UE4SSProgram.cpp
@@ -1059,12 +1059,12 @@ namespace RC
 
     auto UE4SSProgram::install_cpp_mods() -> void
     {
-        install_mods<CppMod>(m_mods);
+        install_mods<CppMod>(get_program().m_mods);
     }
 
     auto UE4SSProgram::install_lua_mods() -> void
     {
-        install_mods<LuaMod>(m_mods);
+        install_mods<LuaMod>(get_program().m_mods);
     }
 
     auto UE4SSProgram::fire_unreal_init_for_cpp_mods() -> void
@@ -1515,7 +1515,7 @@ namespace RC
     auto UE4SSProgram::find_mod_by_name_internal(std::wstring_view mod_name, IsInstalled is_installed, IsStarted is_started, FMBNI_ExtraPredicate extra_predicate)
             -> Mod*
     {
-        auto mod_exists_with_name = std::find_if(m_mods.begin(), m_mods.end(), [&](auto& elem) -> bool {
+        auto mod_exists_with_name = std::find_if(get_program().m_mods.begin(), get_program().m_mods.end(), [&](auto& elem) -> bool {
             bool found = true;
 
             if (!extra_predicate(elem.get()))
@@ -1539,7 +1539,7 @@ namespace RC
         });
 
         // clang-format off
-        if (mod_exists_with_name == m_mods.end())
+        if (mod_exists_with_name == get_program().m_mods.end())
         {
             return nullptr;
         }


deps/first/DynamicOutput/include/DynamicOutput/FileDevice.hpp
```diff
@@ -58,7 +58,7 @@ namespace RC::Output
     {
         if (m_always_create_file)
         {
-            m_file = File::open(m_file_name_and_path, File::OpenFor::Appending, File::OverwriteExistingFile::Yes, File::CreateIfNonExistent::Yes);
+            m_file = File::open(m_file_name_and_path, File::OpenFor::Writing, File::OverwriteExistingFile::Yes, File::CreateIfNonExistent::Yes);
```
UE4SS, line 61: We actually do want to append though.

Yangff: hmm.. in that case you should just use `m_always_create_file=false`, which will append also create if not exists..
Append + OverwriteExisting makes no sense.

UE4SS: The `Append` happens when the file gets written to, and the `OverwriteExisting` happens when `File` is constructed via `File::open`.
We want to overwrite the file if it exists, and after that we want to append to the file every time anything is written to it.

Yangff: When you open a file with `FILE_APPEND_DATA`, it sets the file position to the end of the file (pos = end of file). However, when you also specify `OverwriteExisting = true`, it truncates the file to 0, which effectively sets the position to the beginning (pos = 0). This setup has no impact on the writing process itself, as it will always write to the end in both scenarios.

And, writing operations will always append to the end of the file in both cases. even if you set to `OpenFor::Writing` so at least one of them is wrong...

UE4SS: A file can be written to more than once without being closed and reopened which means it won't get overwritten.

This is what the [Win32 API](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-writefile) says:

> To write to the end of file, specify both the Offset and OffsetHigh members of the OVERLAPPED structure as 0xFFFFFFFF. This is functionally equivalent to previously calling the CreateFile function to open hFile using FILE_APPEND_DATA access.

So according to that, `OpenFor::Writing` would not append to the file because `OpenFor::Writing` uses `GENERIC_WRITE` not `FILE_APPEND_DATA` on Windows.

Yangff: But you're using OverwriteExisting = true at the same  time which means the file will be cleared. So it still won't append.

And, because lpOverlapped is just `nullptr` https://github.com/UE4SS-RE/RE-UE4SS/blob/main/deps/first/File/src/FileType/WinFile.cpp#L99 and you're not using `FILE_FLAG_OVERLAPPED ` flag , effectively it will just append to the end everytime WriteFile is called.

And overlapped is something that allows you to write to the same file object asynchronously and that is why you need to specify the range for it to work (for example you can issue multiple write from multiple parts of memory), and also the meaning of setting offsets to 0xffffffff (to write to the end). And this has nothing to do with the normal behaviour of the `WriteFile`

The following should applies to the normal case:
> Considerations for working with synchronous file handles:
If **lpOverlapped is NULL**, the write operation starts at the **current file position** and WriteFile **does not return(( until the operation is complete, and the system updates the file pointer before WriteFile returns.
If lpOverlapped is not NULL, the write operation starts at the offset that is specified in the [OVERLAPPED](https://learn.microsoft.com/en-us/windows/desktop/api/minwinbase/ns-minwinbase-overlapped) structure and WriteFile does not return until the write operation is complete. The system updates the OVERLAPPED Internal and InternalHigh fields and the file pointer before WriteFile returns.

Current file position just increases everytime you write to the file.

UE4SS: We don't want to append after the file is closed, that's not the goal of `NewFileDevice`.
We only want to append while the file is open.
This is what keeps the log file from growing indefinitely.

I don't see why this change is needed.
We clearly want to append, so we're using the append option.
It doesn't matter if it's redundant on Windows, it still logically makes sense to use append instead of write.
There's nothing wrong with this logic that I can see, and changing from append to write just makes it less clear what we're doing.

Yangff: Because for a file, write is enough to make it append to the end indefinitely. Append when open hints that you want to keep whatever in that file and append to the end but overwrite means you want to discard the old data..

UE4SS: > Because for a file, write is enough to make it append to the end indefinitely. So using an append when open a file hints that you want to keep whatever in that file and append to the end of existing data but overwrite means you want to discard the old data..

Where is this definition of append ?
To me, appending to a file just means append something to the file, it doesn't imply anything about the lifetime of what you're appending to.

I can see how the name `OpenFor` could potentially be confusing though, and it can only be a good thing to create better documentation for what that param actually does and the fact that it operates fully independently of whether you're overwriting the file.

Yangff: Hmm.. idk maybe that's just me feels weird about it..?

UE4SS:
I've gotten rid of all of the unnecessary commits, including the one introducing the commit changing`Appending` => `Writing`.
This commit is now ready for review.

UE4SS (approved):
I think this is a reasonable change, and the code looks good.

UE4SS:
Note that I don't think this is a reasonable change based on the reason provided by this PR, but simply because statics are bad and should be avoided, and I've actually run into several problems related to statics recently and I'd like to remove as many of them as possible from our code.

PR #568, opened 2024-06-15
Use polyhook with xmake package paradigm instead of custom cmakelists.txt

bitonality:
Remove the custom logic in polyhook package inclusion. This unifies xmake packages dir for stable caching on CI runners and polyhook will now respect the xmake package dir setting.

- Adds a package level setting for whether or not to use the zydis submodule that's included in the polyhook2 repo
- Adds package settings for shared/static runtimes
- Adds the ability to reuse the zydis package within polyhook (this also respects zydis AND zycore dependency trees and versioning)


localcc (approved):

bitonality:
Heads up, I want testing on some more machines before merging so please don't merge!

narknon:
Resulted in proxy being unable to load the ue4ss DLL

PR #574, opened 2024-06-18
Port/strings Typedef

Yangff:
This PR just moves `StringType` from File to String.

Right now, CharType is set to `whcar_t`, for a verification build for Linux, one can try to build when it is set to `char16_t`.

It is not intended to solve the circlular dependency in this PR but we may someday as suggested by @localcc seperate Unreal module into Unreal and UnrealDef.

Once this is sovled, CharType should be set to `TCHAR` for normal build.

I think it's safe to say TCHAR will be UTF-16 at this point, if we want to support cases that TCHAR can be say... utf8/utf32.. then might need to wrap it with `to_string` or `to_wstring` based on platform/env.. or pull back the `to_system` macro for those cases..

I have not tested this change but cherrypick it from my current working branch as suggest by @narknon to allow this part to be easily reviewed and merged. I think this should be mostly okay but please check if it compiles.. 
In theory, it should not break anything because I re-exported everything under the `RC::File::` namespace.

It is based on the `fmt` PR although it may not actaully related. It is also possible to merge this before the fmt change.

These is nothing to document at this point, although future code should avoid use `File::*` for string/char type I think.

Also, please try to avoid the use of wstring unless needed.

### DIFF
--- a/deps/first/File/include/File/Macros.hpp
+++ b/deps/first/File/include/File/Macros.hpp
@@ -1,16 +1,5 @@
 #pragma once
 
-// Set this to 1 to use ANSI (char*) instead of wide strings (wchar_t*)
-#ifndef RC_IS_ANSI
-#define RC_IS_ANSI 0
-#endif
-
-#if RC_IS_ANSI == 1
-#define STR(str) u##str
-#else
-#define STR(str) L##str
-#endif
-
 #define THROW_INTERNAL_FILE_ERROR(msg)                                                                                                                         \
     RC::File::Internal::StaticStorage::internal_error = true;                                                                                                  \
     throw std::runtime_error{msg};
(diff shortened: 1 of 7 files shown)


Yangff:
This depends on #565

narknon:
Requires 
```
includes("String")
```
in deps\first\xmake.lua to build.

narknon:
Also breaks anything referencing the old File ToString 
image

And requires the Unreal FMT branch to be merged.

Yangff:
got it.. will steal some commits from another PR..

narknon:
I merged fmt here and on uepseudo so these could be rebased onto main.

Yangff:
image
I think this should build ... at least on my local.

UE4SS (approved):
Builds successfully for me, and the code looks good.

Buckminsterfullerene02:
Both shipping and debug works on DRG with various tool functions, except debug crashes on UHT dump, but it also does in latest main commit, so it is not related to this PR and I will make a seperate issue for that.

PR #575, opened 2024-06-18
Port/strings fmt

Yangff:
This PR depends on 

- [x] https://github.com/Re-UE4SS/UEPseudo/pull/96 
- [x] https://github.com/UE4SS-RE/RE-UE4SS/pull/565 
- [x] https://github.com/UE4SS-RE/RE-UE4SS/pull/574

Will update once they get merged. (#565 merged, rebased)

But since there is almost no structure change to existing code, only a ~~`to_ue`~~`ensure_str` function is introduced to allow easy conversion from `std::string` world to `std::basic_string<TCHAR>`.

~~I will extend this to allow `std::filesystem::path` also, right now it is using `to_ue(path.string())` which has an indrect to UTF-8 to avoid using `wstring()` in normal code.~~
Done

Although this has less changes but it may ended with more codecvt as far as I understand.

Some APIs are changed to use `std::filesystem::path` to proactively avoid using UTF-16 for file path in Linux.

Some windows only strings and literal will keep using wstring and `L""` this includes some use of them in the UVTD. Also I changed some of the generated dllmain in that way. 

Defining the `FORCE_U16` macro we can examine if this system to be compatible with Linux on Windows. It takes advantage of the fact that `char16_t` and `wchat_t` cannot be implicitly converted although they are the same size. However, this means I have to insert some `(TCHAR*)x` and `(CharType*)y` type conversions. Under normal circumstances they won't make any difference. I think it's worth it because it lowers the bar for checking compatibility.


Yangff:
Should be almost done here... should we again apply `to_ue` automatically on `Output::send`? It is about 22 occurances in current code.. And is `to_ue` the correct name?

UE4SS:
> Should be almost done here... should we again apply `to_ue` automatically on `Output::send`? It is about 22 occurances in current code.. And is `to_ue` the correct name?

The name `to_ue` is certainly incorrect if it's supposed to be used in conjunction with `Output::send` internally, because `Output::send` is not related to UE in any way, it's the way we output messages to the system, so it should be a system string shouldn't it ?

Yangff:
> > Should be almost done here... should we again apply `to_ue` automatically on `Output::send`? It is about 22 occurances in current code.. And is `to_ue` the correct name?
> 
> The name `to_ue` is certainly incorrect if it's supposed to be used in conjunction with `Output::send` internally, because `Output::send` is not related to UE in any way, it's the way we output messages to the system, so it should be a system string shouldn't it ?

It used to be, but now every strings we used is ue string unless it's directly involved with lua or the OS (including c style standard libs..)

UE4SS:
> > > Should be almost done here... should we again apply `to_ue` automatically on `Output::send`? It is about 22 occurances in current code.. And is `to_ue` the correct name?
> > 
> > 
> > The name `to_ue` is certainly incorrect if it's supposed to be used in conjunction with `Output::send` internally, because `Output::send` is not related to UE in any way, it's the way we output messages to the system, so it should be a system string shouldn't it ?
> 
> It used to be, but now every strings we used is ue string (or just call it StringType) unless it's directly involved with lua or the OS (including c style standard libs..)

Well since it appears that `to_ue` acts as some kind of magic "make me work" function for strings, there has to be a more appropriate name for it.
Maybe `ensure_str(...)` or something, since it ensures that the string is of the correct type, by conversion when necessary, although I don't like that `ensure` doesn't imply the possibility a conversion, but I think it's better than `to_ue` at least.
Possibly `to_generic(...)` ? If we still want to imply some kind of type conversion.
Any other suggestions are welcome obviously.

BlazesRus:
Maybe rename to to_supportedString?

BlazesRus:
Maybe toSupportedStr then?

BlazesRus:
Or toUEString

UE4SS:
> Maybe toSupportedStr then?

We use snake_case, so the shortest version of your suggestion would be `to_supported_str`.



> Or toUEString

We should avoid mentioning UE for this function as mentioned [here](https://github.com/UE4SS-RE/RE-UE4SS/pull/575#issuecomment-2178016560).

UE4SS:
> This combined with the UEPseudo#96 builds. However, running it gives this error: `Error: exists: An illegal character was encountered. For a multi-byte character set this includes a lead byte without a succeeding trail byte. For the Unicode character set this includes the characters 0xFFFF and 0xFFFE.: "?????????????????????????????????????????????????????????????????????????????\mods.txt"`

I believe this is because `UE4SSProgram::get_mods_directory()`, alongside all of our other string getters, returns a StringView but passes a filesystem::path to `ensure_str` which will get converted to a temporary StringType and returned, thus leaving us with a view of a temporary that's gone out of scope.

I will see if we can change our getters to return filesystem::path instead of a string.

UE4SS:
> > This combined with the UEPseudo#96 builds. However, running it gives this error: `Error: exists: An illegal character was encountered. For a multi-byte character set this includes a lead byte without a succeeding trail byte. For the Unicode character set this includes the characters 0xFFFF and 0xFFFE.: "?????????????????????????????????????????????????????????????????????????????\mods.txt"`
> 
> I believe this is because `UE4SSProgram::get_mods_directory()`, alongside all of our other string getters, returns a StringView but passes a filesystem::path to `ensure_str` which will get converted to a temporary StringType and returned, thus leaving us with a view of a temporary that's gone out of scope.
> 
> I will see if we can change our getters to return filesystem::path instead of a string.

On second thought, I think it's better to return `StringType` instead, it's not as efficient but at least way we maintain compatibility with everyones code this way.

UE4SS:
> This combined with the UEPseudo#96 builds. However, running it gives this error: `Error: exists: An illegal character was encountered. For a multi-byte character set this includes a lead byte without a succeeding trail byte. For the Unicode character set this includes the characters 0xFFFF and 0xFFFE.: "?????????????????????????????????????????????????????????????????????????????\mods.txt"`

This is now fixed, and the code looks good to me.
If anyone else wants to review it, go for it, otherwise I think we're ready as soon as https://github.com/Re-UE4SS/UEPseudo/pull/96/ is ready.

UE4SS/include/SDKGenerator/UEHeaderGenerator.hpp
```diff
@@ -91,21 +93,21 @@ namespace RC::UEGenerator
 
     struct StringInsensitiveCompare
     {
-        auto operator()(const std::wstring& a, const std::wstring& b) const -> bool
+        auto operator()(const StringType& a, const StringType& b) const -> bool
         {
-            return _wcsicmp(a.c_str(), b.c_str()) < 0;
+            return RC::Unreal::FGenericPlatformStricmp::Stricmp((RC::Unreal::UTF16CHAR*) a.c_str(), (RC::Unreal::UTF16CHAR*)b .c_str()) < 0;
```
narknon, line 98: Some weird formatting here.

UE4SS/src/GUI/Dumpers.cpp
```diff
@@ -83,21 +83,21 @@ namespace RC::GUI::Dumpers
     {
         StringType root_actor_buffer{};
 
-        static auto location_property = root_component->GetPropertyByNameInChain(STR("RelativeLocation"));
-        static auto rotation_property = root_component->GetPropertyByNameInChain(STR("RelativeRotation"));
-        static auto scale_property = root_component->GetPropertyByNameInChain(STR("RelativeScale3D"));
+        static auto location_property = root_component->GetPropertyByNameInChain(FromCharTypePtr<TCHAR>(STR("RelativeLocation")));
```
narknon, line 86: We should have documentation on when to use this cast.

narknon (changes requested):

UE4SS/include/SDKGenerator/UEHeaderGenerator.hpp
UE4SS: Doesn't look like weird formatting to me, but it does use two C-casts which is bad, an I don't think it's even correct ?
The `Stricmp` function has overloads for all the string types so I'm not sure why we need to cast or convert them.
I will remove these casts, and someone will have to test that UHT is still working properly after this.

UE4SS/src/GUI/Dumpers.cpp
UE4SS: So this function is for testing purposes only until Linux support has been merged, and after that the function can be removed.

See this and the comment below it: https://github.com/Re-UE4SS/UEPseudo/pull/96#issuecomment-2174994211

localcc: As I understand, `FromCharTypePtr` should be used now for whenever we convert a char pointer to unreal character pointer. And this function will only work if those are the same size so no conversion would be needed. And this would be removed by the time linux support gets introduced?

UE4SS: > As I understand, `FromCharTypePtr` should be used now for whenever we convert a char pointer to unreal character pointer. And this function will only work if those are the same size so no conversion would be needed. And this would be removed by the time linux support gets introduced?

Yes, and it would likely be removed either by the linux support PR (at which point CI takes over responsibility to make sure the Linux build doesn't break) or soon after if it gets forgotten.

UE4SS/src/SDKGenerator/Common.cpp
```diff
@@ -123,14 +123,14 @@ namespace RC::UEGenerator
 
     auto generate_delegate_name(FProperty* property, const File::StringType& context_name) -> File::StringType
     {
-        const std::wstring property_name = sanitize_property_name(property->GetName());
+        const auto property_name = sanitize_property_name(property->GetName());
```
narknon, line 126: Not a huge fan of using auto in this and the following contexts.

UE4SS: Why not ?

narknon: Because at a glance you can't tell what type GetName returns.  In this context based on the name of the function it's likely some form of string, but we can't really make that assumption with "Name" given FNames.
```const StringType property_name = sanitize_property_name(property->GetName());``` would be preferable

UE4SS: I would argue that you don't need to know what type it returns.
I don't particularly care myself if auto is used here or not, but I don't think the creator of this PR will respond regardless so if you want this change, then you're probably gonna have to make it yourself.

narknon (approved):

PR #591, opened 2024-07-05
Add CI builds for pull requests

bitonality:
This adds xmake building to Pull Requests whenever they are created or updated. This should ensure that we aren't breaking compilation when checking in new code. The check is NOT mandatory for now due to the "partner PR" not being implemented between UEPseudo and UE4SS, so there will be some PR completions that don't compile, but shouldn't necessarily block a merge. Long term I plan on implementing partner PRs between the two repos so we can turn on enforcement before check-in, but this should suffice for now.

#### Build All Modes Workflow

This workflow automates the build process for all modes of UE4SS whenever a pull request is made to the main branch, excluding changes to documentation files. It dynamically calculates the build matrix based on the modes defined in the project and reports artifacts to the PR comments. Key features include:

1. **Concurrency Control**: Ensures only the latest commit triggers the workflow, preventing multiple builds for rapid successive pushes.
2. **Dynamic Mode Calculation**: Calculates which modes to build and which should upload artifacts based on project configurations.
3. **Build Execution**: Uses a reusable workflow to build each mode determined in the matrix, ensuring consistent builds across different configurations.
4. **Artifact Reporting**: Uploads build artifacts and reports them as comments on the pull request, providing clear visibility of the build results.

#### Build UE4SS Workflow

This reusable workflow handles the installation of xmake, cloning of the repository, building of UE4SS, and uploading of artifacts. It provides a consistent environment for any future workflows or tasks requiring UE4SS to be built. Key features include:

1. **Repository Checkout**: Clones the repository and its submodules for full project context.
2. **Cache Management**: Utilizes caching to optimize build times by storing dependencies and using a weekly cache key.
3. **Environment Setup**: Installs necessary tools, such as Visual Studio Build Tools and xmake, to configure the build environment.
4. **Dependency Handling**: Restores and saves cached dependencies to reduce redundant installations.
5. **Build Execution**: Configures and builds the project using xmake, tailored to the specified build mode.
6. **Artifact Management**: Calculates and uploads build artifacts based on specified targets, retaining them for a defined period.

These workflows ensure efficient, consistent, and automated builds for the UE4SS project, enhancing development productivity and reliability.


Buckminsterfullerene02:
I think it has been more than long enough for anyone else to make any rejections so I'll just go ahead and merge this and we'll see how it goes.

PR #617, opened 2024-08-11
Version checking for crash dumper

Buckminsterfullerene02:
**Description**

We keep getting issues posted where the user is using an old version of UE4SS that does not work for their game. 

Yangff suggested adding information that checks for new versions when crash. 

Currently this change is adding the information in the message box that opens when it makes a crash dump, however the code is exposed in UE4SSProgram.hpp so could be used in other places such as in the logs or somewhere more obvious to the user, and it seems useful enough to add in the mod APIs in the future. 

**Type of change**

**How Has This Been Tested?**

Tested output of warning by temporarily flipping `if (!is_latest_ue4ss_version(latest_version))` condition. 

Checked that config works as intended.

Need to force crash dump to happen to test crash dumper part.

**Checklist**

### DIFF
--- a/UE4SS/include/SettingsManager.hpp
+++ b/UE4SS/include/SettingsManager.hpp
@@ -25,6 +25,7 @@ namespace RC
             bool EnableDebugKeyBindings{false};
             int64_t SecondsToScanBeforeGivingUp{30};
             bool UseUObjectArrayCache{true};
+            bool LatestVersionCheck{true};
         } General;
 
         struct SectionEngineVersionOverride
--- a/UE4SS/include/UE4SSProgram.hpp
+++ b/UE4SS/include/UE4SSProgram.hpp
@@ -221,6 +221,9 @@ namespace RC
         RC_UE4SS_API auto generate_uht_compatible_headers() -> void;
         RC_UE4SS_API auto generate_cxx_headers(const std::filesystem::path& output_dir) -> void;
         RC_UE4SS_API auto generate_lua_types(const std::filesystem::path& output_dir) -> void;
+        RC_UE4SS_API auto get_latest_version_check_setting() -> bool;
+        RC_UE4SS_API auto get_latest_ue4ss_version() -> StringType;
+        RC_UE4SS_API auto is_latest_ue4ss_version(StringType latest_ver) -> bool;
         auto get_debugging_ui() -> GUI::DebuggingGUI&
         {
             return m_debugging_gui;
--- a/UE4SS/src/CrashDumper.cpp
+++ b/UE4SS/src/CrashDumper.cpp
@@ -60,8 +60,18 @@ namespace RC
             return EXCEPTION_CONTINUE_SEARCH;
         }
 
-        const StringType message = fmt::format(STR("Crashdump written to: {}"), dump_path);
-        MessageBoxW(NULL, FromCharTypePtr<wchar_t>(message.c_str()), L"Fatal Error!", MB_OK);
+        StringType version_message = L"";
+        if (UE4SSProgram::get_program().get_latest_version_check_setting())
+        {
+            StringType latest_version = UE4SSProgram::get_program().get_latest_ue4ss_version();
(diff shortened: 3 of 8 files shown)


Buckminsterfullerene02:
Draft PR because code currently does not build due to a bunch of seemingly unrelated errors in LiveView.cpp and Dumpers.cpp. @UE4SS any ideas?

UE4SS:
> Draft PR because code currently does not build due to a bunch of seemingly unrelated errors in LiveView.cpp and Dumpers.cpp. @UE4SS any ideas?

This is caused by `curl.h` either defining the `max` macro or including something that does.
Perhaps it leaks `Windows.h` which is usually the root of this problem.
The solution is to move the `curl.h` include to the cpp file.

UE4SS:
Just fyi, you don't need to put all functions in the header.
Unless you want to call the function outside `UE4SSProgram.cpp`, or you want other people to be able to call the function, it shouldn't go in the header because it's more difficult to find what you're looking for if the header is big and bloated, and this is especially problematic if it's an exported header meant for other people (mods) to use.
The `is` and `get` functions make sense to put in the header, but the rest of them seem like internal details that no one is going to need to know exists.
Static free functions in the .cpp file is what should be used for internal functions unless they absolutely need private access to a class.

It's not a blocking problem, but it's something to keep in mind, at least for future PRs.

Buckminsterfullerene02:
Need to force crash dump to happen to test crash dumper part. I tried dereferincing null ptr, accessing index out of range of array and infinite recursion, and none of them cause the crash (I put breakpoint on and they are running past that code) so I am out of ideas of ways to force a crash. I swear I've done this before successfully...

Buckminsterfullerene02:
> Just fyi, you don't need to put all functions in the header. Unless you want to call the function outside `UE4SSProgram.cpp`, or you want other people to be able to call the function, it shouldn't go in the header because it's more difficult to find what you're looking for if the header is big and bloated, and this is especially problematic if it's an exported header meant for other people (mods) to use. The `is` and `get` functions make sense to put in the header, but the rest of them seem like internal details that no one is going to need to know exists. Static free functions in the .cpp file is what should be used for internal functions unless they absolutely need private access to a class.
> 
> It's not a blocking problem, but it's something to keep in mind, at least for future PRs.

I wasn't even aware you could do this, mind showing an example? Or doing it with `write_callback` or `split_version` yourself to show me for as example.

igromanru:
Personally, I find the feature unnecessary.  Isn't it the first feature that requires an internet connection?  
When a single game suddenly asks for internet permissions, I'm always suspicious and I have to look in the firewall what it actually does. That's why I'm not a fan of such things, unless it's necessary.  

> We keep getting issues posted where the user is using an old version of UE4SS that does not work for their game.

At current state a lot of games require the latest experimental version to work, in such case it's not helpful. Also a lot of users don't even read the log themself, they just post "it doesn't work with game XYZ".  
If they aren't even smart enough to try out the latest Release version before posting an issue nothing will help.

UE4SS:
> Personally, I find the feature unnecessary. Isn't it the first feature that requires an internet connection? When a single game suddenly asks for internet permissions, I'm always suspicious and I have to look in the firewall what it actually does. That's why I'm not a fan of such things, unless it's necessary.
> 
> > We keep getting issues posted where the user is using an old version of UE4SS that does not work for their game.
> 
> At current state a lot of games require the latest experimental version to work, in such case it's not helpful. Also a lot of users don't even read the log themself, they just post "it doesn't work with game XYZ". If they aren't even smart enough to try out the latest Release version before posting an issue nothing will help.

I think your points here are very valid, and I agree to some extent.
I don't think we should check for a new version on startup, because no one checks the log and a message box would just get in the way, but I think checking when it crashes and displaying it in the preexisting error message box is fine.
Currently it checks on startup and when it crashes.

The feature itself obviously requires an internet connection, but if there isn't a connection, as far as I know it will currently fall to the "a new version is available" case, which isn't great and it would be good if this was changed if this PR goes forward.

Issue #615, opened 2024-08-11
[BUG - Release] Unable to Build Latest Main at time of writing.


portalmaster137:
I've been trying to build the latest main build (07bb2dd), and no matter what i do, it keeps failing to build.

Here are my commands in order, with the xmake log attached.
```
git clone git@github.com:UE4SS-RE/RE-UE4SS.git
cd .\RE-UE4SS\
git submodule update --init --recursive
xmake f -m "Game__Shipping__Win64"
xmake -vD
```
I Beleive this is an error with one of the submodules (the UEPseudo), as i cant get that to build separately either, i've also tried for hours to get this work but i've had no luck, so people smarter than me can hopefully fix this issue.

[xmake_log.txt](https://github.com/user-attachments/files/16571303/xmake_log.txt)

UE4SS:
I don't have a solution for you other than defining `EAspectRatioAxisConstraint` yourself somewhere.
The underlying problem is UVTD generating forward declarations for enums when that's prohibited in C++.
The solution is to fix UVTD.
That said, no one else seems to be having this problem so clearly MSVC has implemented some kind of compiler extension to allow this behavior or it's a compiler bug.
Regardless, I don't know why you have this problem and no one else.

portalmaster137:
Would it be worth it to reinstall visual studio from the ground up? Could it also be possible that 2022 Preview is somehow messing it up?

portalmaster137:
Found the problem!
It was MSVC Itself, 19.41, aka Visual Studio 2022 Preview Would NOT Compile, however 19.40, which is used in normal VS2022, works just fine. I'm thinking MSVC 19.41 Disables forward decleration of enums or somthing, but it builds now. Maybe make this a check in xmake or put in the readme to use 19.40?

PR #618, opened 2024-08-11
Add warning for MSVC 19.41 Compiler bug.

portalmaster137:
**Description**

This is a simple warning, since MSVC 19.41 has introduced some kind of compiler bug. #615 
The "Fix" Is to simply not use Visual Studio 2022 Preview for the time being until the UVTD Enums can compile with the new MSVC Version.

**Type of change**

- [x] Other... Please describe: As this is not actually a "Fix" for now, this is just a warning for others to save the headache of figuring out why it wont compile like i did.

**How Has This Been Tested?**

Latest Main could compile fine with MSVC 19.40.33813, which is in VS2022
Using VS2022 Preview, MSVC 19.41.34117 would complain about some forward Enum Declerations in UVTD.

**Checklist**


Buckminsterfullerene02:
Replaced by #619

PR #619, opened 2024-08-12
Raise error or warning depending on installed rust/msvc versions

Buckminsterfullerene02:
**Description**

Some build issues have been caused in the past and recently when the user does not meet the minimum build requirements. 

This PR aims to solve this by checking the Rust and MSVC minimum requirements, and, if they are not met, error out on config, not allowing build to begin in the first place. xmake already uses `set_xmakever` which works fine for minimum version.

In the case of #615, the user is using the current preview version which has a higher MSVC version, and causes the build to fail. This PR additionally prints a warning (and continues build) if they are using this current MSVC version. We should keep an eye on this version and fix the issue at the root or see if this is just a compiler bug, but in either case, eventually remove this temporary measure.

The PR also prints the additional versions info in console without verbose mode needing to be enabled.

Currently hardcoding the minimum versions at the top of the file, but perhaps we could change this to something less terrible in the future.

**Type of change**

- [x] New feature (non-breaking change which adds functionality): xmake rules

**How Has This Been Tested?**

Ran `xmake l detect.sdks.find_vstudio` to get the msvc version strings.

Changed minimum versions for rust and msvc to test edge cases on version numbers, such as equal, or equal majors but different minors/patches, lower and higher than installed version etc. 

**Checklist**

Tests with changed minimums above my installed versions:

image

image

image

Works on any part touching UE4SS configurations:
image

### DIFF
--- a/.github/workflows/msvc_build_ue4ss.yml
+++ b/.github/workflows/msvc_build_ue4ss.yml
@@ -56,6 +56,9 @@ jobs:
       - name: Install VS2022 BuildTools 17.9.7
         run: choco install -y visualstudio2022buildtools --version=117.9.7.0 --params "--add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 --installChannelUri https://aka.ms/vs/17/release/180911598_-255012421/channel"
 
+      - name: Install JQ
+        run: choco install -y jq
+
       - name: Setup xmake
         uses: xmake-io/github-action-setup-xmake@v1
         with:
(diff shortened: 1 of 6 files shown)


UE4SS:
Is it possible to get an override flag for this so that testing new versions doesn't require modifying any xmake files ?
Like `xmake f --allow-all-versions -m "Game__Debug__Win64 -y` or something.

Buckminsterfullerene02 (commented):
@localcc Somehow this change is now ignoring the default value of allowAllVersions being false. 

Meaning that if you run config without the flag, it will just assume that the option is true and skip checking, which is incorrect behaviour.

Buckminsterfullerene02:
Nevermind, I realised that once you set the flag, the option value is set until you specify otherwise, which in hindsight does seem like the correct behaviour when setting config options. 

I did still change the flag to `versionCheck` because the naming makes more sense since you have to pass in `y/n`, and the default is true.

localcc (dismissed):

localcc (approved):

PR #630, opened 2024-08-22
Fix ScriptArray/TArray Lua param pushes

narknon:


Buckminsterfullerene02:
Can you provide an example you used to test this?

narknon:
There are literally no reflected functions taking TArray as an input/non-out param, so there is no example possible.  Confirming it is the only thing that works in Wukong though, so merging.

PR #636, opened 2024-08-27
Port/strings fmt (replaces #575)

narknon:
This PR replaces #575 for CI reasons.

By @Yangff 

This PR depends on 

- [x] https://github.com/Re-UE4SS/UEPseudo/pull/96 
- [x] https://github.com/UE4SS-RE/RE-UE4SS/pull/565 
- [x] https://github.com/UE4SS-RE/RE-UE4SS/pull/574

Will update once they get merged. (#565 merged, rebased)

But since there is almost no structure change to existing code, only a ~~`to_ue`~~`ensure_str` function is introduced to allow easy conversion from `std::string` world to `std::basic_string<TCHAR>`.

~~I will extend this to allow `std::filesystem::path` also, right now it is using `to_ue(path.string())` which has an indrect to UTF-8 to avoid using `wstring()` in normal code.~~
Done

Although this has less changes but it may ended with more codecvt as far as I understand.

Some APIs are changed to use `std::filesystem::path` to proactively avoid using UTF-16 for file path in Linux.

Some windows only strings and literal will keep using wstring and `L""` this includes some use of them in the UVTD. Also I changed some of the generated dllmain in that way. 

Defining the `FORCE_U16` macro we can examine if this system to be compatible with Linux on Windows. It takes advantage of the fact that `char16_t` and `wchat_t` cannot be implicitly converted although they are the same size. However, this means I have to insert some `(TCHAR*)x` and `(CharType*)y` type conversions. Under normal circumstances they won't make any difference. I think it's worth it because it lowers the bar for checking compatibility.

### DIFF
--- a/UE4SS/include/ExceptionHandling.hpp
+++ b/UE4SS/include/ExceptionHandling.hpp
@@ -8,7 +8,7 @@
 #define UE4SS_ERROR_OUTPUTTER()                                                                                                                                \
     if (!Output::has_internal_error())                                                                                                                         \
     {                                                                                                                                                          \
-        Output::send<LogLevel::Error>(STR("Error: {}\n"), to_wstring(e.what()));                                                                               \
+        Output::send<LogLevel::Error>(STR("Error: {}\n"), ensure_str(e.what()));                                                                                    \
     }                                                                                                                                                          \
     else                                                                                                                                                       \
     {                                                                                                                                                          \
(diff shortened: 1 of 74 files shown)


UE4SS (approved):

UE4SS (approved):

narknon:
Crashes when sending strings to UE

PR #239, opened 2023-10-26
Replace polyhook with safetyhook

localcc:
This greatly simplifies code and reduces boilerplate

PR #649, opened 2024-09-06
Polyhook -> safetyhook

narknon:


UE4SS:
This PR completely removes PolyHook, and that will force mod creators that utilize PolyHook to update their code or re-add PolyHook themselves, this needs to be mentioned in the changelog.

igromanru:
Wait a minute.
Is it the same as #239?

PR #664, opened 2024-09-26
Implement global Lua function GenerateLuaTypes (#625)

### LINKED ISSUE #625: [Feature] Register "UE4SSProgram::generate_lua_types" as a lua function
So far you can call all the dumper functions through lua. Like "DumpAllObjects", "GenerateSDK", etc.
But the [generate_lua_types](https://github.com/UE4SS-RE/RE-UE4SS/blob/main/UE4SS/src/UE4SSProgram.cpp#L1433) is missing.
It should be possible to register it here similar to other functions:
https://github.com/UE4SS-RE/RE-UE4SS/blob/main/UE4SS/src/Mod/LuaMod.cpp#L1434

igromanru:
Added  GenerateLuaTypes like described in #625.
Additionally added dumpers functions to Types.lua.

Built UE4SS and tested in Abiotic Factor. The function works and does exactly the same as the button in the GUI.

resolve #625

### DIFF
--- a/UE4SS/src/Mod/LuaMod.cpp
+++ b/UE4SS/src/Mod/LuaMod.cpp
@@ -1435,7 +1435,7 @@ Overloads:
                 const Mod* mod = get_mod_ref(lua);
                 if (!mod)
                 {
-                    lua.throw_error("Could not dump objects and properties because the pointer to 'Mod' was nullptr");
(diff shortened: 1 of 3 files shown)


UE4SS (dismissed):
The code looks good, but I haven't tested it.

igromanru:
Conflict resolved

UE4SS (approved):

PR #666, opened 2024-09-27
Added global lua function CreateInvalidObject (#652)

### LINKED ISSUE #652: [Feature Request] Add new() constructor for RemoteObject/LocalObject or a Global Function that can create an invalid object
UE4SS is currently set up in such a way that it uses weak pointer objects everywhere aka. RemoteObject/LocalObject which are objects that always exist but have an `IsValid` function to check if the pointer that points to the object is valid or simply a placeholder that always returns false.
It would be good to give Lua Mod developers a way to develop their mods with the same approach in mind that each object isn't nil and can be checked with `IsValid`.
Therefore I suggest to a typical `:new()` method for RemoteObject/LocalObject that can be used by Lua mod devs to create new empty/invalid objects instead of dealing with `nil` values in their own code.  
If it's easier it can also be just a global function like `CreateEmptyObject`, `CreateRemoteObject`, `CreateInvalidObject` or whatever you want to call it.

igromanru:
Resolves #652

**Testing**
Built UE4SS and deployed in Abiotic Factor.
Tested with Lua code:
```lua
local blankObject = CreateBlankObject()
print("blankObject varibale type: " .. type(blankObject))
if blankObject then
	print("blankObject userdata type: " .. blankObject:type())
	print("blankObject IsValid:" .. tostring(blankObject:IsValid()))
end
```
Result:  
```
[12:40:32.1231285] [Lua] blankObject varibale type: userdata
[12:40:32.1231360] [Lua] blankObject userdata type: UObject
[12:40:32.1231428] [Lua] blankObject IsValid:false
```

**Additional**
Fixed `FindFirstOf` return type annotation in `Types.lua`

**Documentation**
All changes documented in `Changelog.md` and created a new docs page for `CreateBlankObject`.

### DIFF
--- a/assets/Changelog.md
+++ b/assets/Changelog.md
@@ -21,6 +21,7 @@ Added search filter: `IncludeClassNames`. ([UE4SS #472](https://github.com/UE4SS
 ### UHT Dumper
 
 ### Lua API
+Added global function `CreateInvalidObject`, which returns an invalid UObject. ([UE4SS #652](https://github.com/UE4SS-RE/RE-UE4SS/issues/652)) 
 
 ### C++ API
 Key binds created with `UE4SSProgram::register_keydown_event` end up being duplicated upon mod hot-reload.  
@@ -109,6 +110,8 @@ Fixed `FText` not working as a parameter (in `RegisterHook`, etc.) ([UE4SS #422]
 Fixed crash when calling UFunctions that take one or more 'out' params of type TArray<T>. ([UE4SS #477](https://github.com/UE4SS-RE/RE-UE4SS/pull/477))
 
 Fixed `RegisterProcessConsoleExecPostHook`. ([UE4SS #631](https://github.com/UE4SS-RE/RE-UE4SS/pull/631))
+
+Fixed `FindFirstOf` return type annotation in `Types.lua` to signal that the return value will never be nil. ([UE4SS #652](https://github.com/UE4SS-RE/RE-UE4SS/issues/652))
 
 ### C++ API
 Fixed a crash caused by a race condition enabled by C++ mods using `UE4SS_ENABLE_IMGUI` in their constructor ([UE4SS #481](https://github.com/UE4SS-RE/RE-UE4SS/pull/481))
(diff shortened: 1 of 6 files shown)


assets/Changelog.md
```diff
@@ -21,6 +21,7 @@ Added search filter: `IncludeClassNames`. ([UE4SS #472](https://github.com/UE4SS
 ### UHT Dumper
 
 ### Lua API
+Added global function `CreateBlankObject`, which returns an empty UObject. ([UE4SS #652](https://github.com/UE4SS-RE/RE-UE4SS/issues/652))
```
UE4SS, line 24:
```suggestion
Added global function `CreateBlankObject`, which returns an invalid UObject. ([UE4SS #652](https://github.com/UE4SS-RE/RE-UE4SS/issues/652)) 
```
I think this makes more sense.
I don't think "empty" makes it clear what the purpose of this function is.

UE4SS (changes requested):

igromanru:
Can someone just Clone the branch from my fork, compile and test it locally, then just force merge?

UE4SS (approved):

narknon:
I think "Blank" is a bit confusing.  Maybe dummy or empty?

UE4SS:
> I think "Blank" is a bit confusing. Maybe dummy or empty?

Perhaps "CreateInvalidObject" is even more accurate to what the function actually does ?

narknon:
I guess CreateInvalidLuaUObject is too verbose

UE4SS:
> I guess CreateInvalidLuaUObject is too verbose

Yeah I don't think "LuaU" is necessary, and besides, we might reuse this function at some point for non-UObject stuff so it's nice to keep the "U" out of "Object".
It really doesn't have anything to do with UObject anyway, it's just a function implemented for UObjects among other objects.

igromanru:
Tested;
```lua
local invalidObject = CreateInvalidObject()
print("blankObject variable type: " .. type(invalidObject))
if invalidObject then
	print("blankObject userdata type: " .. invalidObject:type())
	print("blankObject IsValid: " .. tostring(invalidObject:IsValid()))
end
```
Result:  
```
blankObject variable type: userdata
blankObject userdata type: UObject
blankObject IsValid: false
```

UE4SS (approved):

PR #682, opened 2024-10-07
Added note that RegisterHook doesn't work with delegate functions

### LINKED ISSUE #483: [BUG - Release] Hooking DelegateFunction does not work (kind of)
**Branch or Release**
3.0.1

**Game and Engine Version**
Tested on grounded 1.4.0.4495 Rel (ue 4.27.2) and palworld 0.2.2.0 (ue 5.1.1)

**Describe the bug**
Using RegisterHook on DelegateFunction does not actually lead to hook execution
As far as I understand, because these are just reflections and are actually called natively?
The example below is generalized, I tested it on larger number of delegates from grounded

**Mods directory**

<details>
<summary>Generalized test mod</summary>

```lua
RegisterHook("/Script/Engine.OnAnimInitialized__DelegateSignature", function()
 print("/Script/Engine.OnAnimInitialized__DelegateSignature \n")
end)
RegisterHook("/Script/Engine.OnAudioPlayStateChanged__DelegateSignature", function()
 print("/Script/Engine.OnAudioPlayStateChanged__DelegateSignature \n")
end)
RegisterHook("/Script/Engine.OnAudioPlaybackPercent__DelegateSignature", function()
 print("/Script/Engine.OnAudioPlaybackPercent__DelegateSignature \n")
end)
RegisterHook("/Script/Engine.InputActionHandlerDynamicSignature__DelegateSignature", function()
 print("/Script/Engine.InputActionHandlerDynamicSignature__DelegateSignature \n")
end)
```

</details>

**To Reproduce**
1. Launch game
2. Trigger any anim/audio component
3. See nothing from the hooks is printed

**Expected behavior**
Hooks are being executed

 - OS: Win11

igromanru:
Resolves #483

Added note that RegisterHook doesn't work with delegate functions and improved the example code

### DIFF
--- a/docs/lua-api/global-functions/registerhook.md
+++ b/docs/lua-api/global-functions/registerhook.md
@@ -10,6 +10,8 @@ Returns two ids, both of which must be passed to `UnregisterHook` if you want to
 
 > Any `UFunction` that you attempt to register with `RegisterHook` must already exist in memory when you register it.  
 
+> `RegisterHook` doesn't support delegate functions!
+
 ## Parameters
 
 | # | Type     | Information                                                                                                                                                         |
@@ -27,7 +29,18 @@ Returns two ids, both of which must be passed to `UnregisterHook` if you want to
 
 ## Example
 ```lua
-RegisterHook("/Script/Engine.PlayerController:ClientRestart", function()
-    print("PlayerController restarted\n")
+PreId, PostId = RegisterHook("/Script/Engine.PlayerController:ClientRestart", function(Context, NewPawn)
+    local playerController = Context:get()
+    local newPawn = NewPawn:get()
+
+    print("PayerController FullName: " .. playerController:GetFullName())
+    if newPawn:IsValid() then
+        print("NewPawn FullName: " .. newPawn:GetFullName())
+    end
+
+    if PreId then
+        -- Unhook once the function has been called
+        UnregisterHook("/Script/Engine.PlayerController:ClientRestart", PreId, PostId)
+    end
 end)
 ```


docs/lua-api/global-functions/registerhook.md
```diff
@@ -27,7 +29,18 @@ Returns two ids, both of which must be passed to `UnregisterHook` if you want to
...
-RegisterHook("/Script/Engine.PlayerController:ClientRestart", function()
-    print("PlayerController restarted\n")
+PreId, PostId = RegisterHook("/Script/Engine.PlayerController:ClientRestart", function(Context, NewPawn)
+    local playerController = Context:get()
+    local newPawn = NewPawn:get()
+
+    print("PayerController FullName: " .. playerController:GetFullName())
+    if newPawn:IsValid() then
+        print("NewPawn FullName: " .. newPawn:GetFullName())
+    end
```
UE4SS, line 39:
```suggestion
PreId, PostId = RegisterHook("/Script/Engine.PlayerController:ClientRestart", function(Context_Param, NewPawn_Param)
    local PlayerController = Context_Param:get()
    local NewPawn = NewPawn_Param:get()

    print("PayerController FullName: " .. PlayerController:GetFullName())
    if NewPawn:IsValid() then
        print("NewPawn FullName: " .. NewPawn:GetFullName())
    end
```
Potential for some confusion + mixed casing.

I've changed the params to use the `_Param` suffix which serves the purpose of telling the reader that this is a Local/RemoteUnrealParam and thus needs `get()` to be called on it, and it also serves to keep the casing consistent.

An alternative to `_Param`, would be the `In<Name>`/`Out<Name>` style, where ` In` is used for input params and `Out` is used for output params, of which there are none in this example to be clear.

igromanru: It bothers me for a while already but I try to keep the code style similar to what it was before.  
In Lua, it is atypical to use PascalCase for local variables and parameters, it throws me off track every time.   

However, while I agree that local PlayerController and NewPawn should probably be PascalCase, I don't agree with adding special prefixes/suffixes to the names of callbacks variables, simply because it's not coherent with all the other examples and code we use in UEHelpers and other standard mods.

UE4SS: We use PascalCase simply because that's what UE uses.
I guess we can keep this PR unchanged and do a larger pass on all the docs later to address this.

igromanru: Unreal Engine is made with modern Microsoft like C++/C# code style.  
In untyped language like lua it's easier to distinguish between global and local variables if they have different casing style.  
While I personally don't like snake case, I use for my own code camelCase for local variables and parameters.

UE4SS:
> Should this close #483 if there's plans in the future to support delegate hooking? (if the issue was meant for tracking)

No, it shouldn't.

igromanru:
> Should this close 483 if there's plans in the future to support delegate hooking? (if the issue was meant for tracking)

The issue is a bug report. RegisterHook not supporting delegates is not a bug. It requires special handing and someone has to invest unknown amount of time to implement it. I don't even know how it can be done and if it should be a part `RegisterHook`. The whole system probably requires a rework to support other things that were on ToDo like PreHooks for Blueprint functions etc.  
Imo, closing a bug report on the grounds that it is not a feature is an acceptable way to keep the amount of issues at bay.  
However, I've removed "Resolves" command from the first post.

UE4SS (approved):

igromanru:
Wait a second, I'm about to add Callback Parameters description

docs/lua-api/global-functions/registerhook.md
```diff
@@ -27,7 +29,18 @@ Returns two ids, both of which must be passed to `UnregisterHook` if you want to
...
-RegisterHook("/Script/Engine.PlayerController:ClientRestart", function()
-    print("PlayerController restarted\n")
+PreId, PostId = RegisterHook("/Script/Engine.PlayerController:ClientRestart", function(Context, NewPawn)
+    local PlayerController = Context:get()
+    local Pawn = NewPawn:get()
+
+    print("PayerController FullName: " .. PlayerController:GetFullName())
+    if Pawn:IsValid() then
+        print("NewPawn FullName: " .. Pawn:GetFullName())
+    end
```
UE4SS, line 39:
```suggestion
    print("PayerController FullName: " .. PlayerController:GetFullName() .. "\n")
    if Pawn:IsValid() then
        print("NewPawn FullName: " .. Pawn:GetFullName() .. "\n")
    end
```

Our `print` function requires an explicit new line.
Our other examples mostly use string.format btw, since you mentioned you wanted to keep to what we already do.

UE4SS (approved):

PR #699, opened 2024-11-01
Fix(Types.lua): Fix GetFName function return definition

0-F:
The return definition of the GetFName function was nil instead of FName.


narknon:
Thank you!

PR #710, opened 2024-12-07
Cross-compile patch & build instructions

UE4SS:
**Description**

Changes to enable cross-compilation with [msvc-wine](https://github.com/mstorsjo/msvc-wine).

**How Has This Been Tested?**

Follow the build instructions, which can be found in README.md included in this pull request.

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,9 @@ Note that you should also commit & push the submodules that you've updated if th
 ### Building Windows binaries on Linux
 
 We only officially support [msvc-wine](https://github.com/mstorsjo/msvc-wine) for cross-compiling.  
+> [!CAUTION]
+> You must use [xmake](https://github.com/xmake-io/xmake) v2.9.7 or later, and as of early December 2024, this version is not yet released which means you must install the dev version of xmake.  
+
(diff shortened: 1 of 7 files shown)


UE4SS:
There were some last minute fixes.
This should be ready to be tested now.

igromanru:
Since you got a lot of `if` there, if it runs in wine, I guess windows shouldn't be affected at all right?  
However, I've successfully compiled the branch in Windows 11.

UE4SS:
> Since you got a lot of `if` there, if it runs in wine, I guess windows shouldn't be affected at all right? However, I've successfully compiled the branch in Windows 11 Pro 23H2.

Yeah, for some reason checking against the default value of the option doesn't work so now we're checking if the option isn't msvc-wine to detect non-cross builds.
CI is passing, and msvc-wine works for me on Arch Linux, so this should be good to go, though it would be good if someone else could test this on Linux just to make sure it's working on more machines than mine.

Buckminsterfullerene02 (approved):
I followed the instructions and was able to easily (aside from one error, which was an easy fix and is now in the PR) build the project on arch linux. 

I tested the compiled binaries on windows and it works. I tested kismet debugger, pressed the various generator buttons and checked live view, all work as expected.

PR #711, opened 2024-12-07
Fix readme file not populating on the correct docs versions

Buckminsterfullerene02:
**Description**

Currently, ue4ss home page and release tag (i.e. v3.0.1) shows the instructions for building ue4ss against main branch, which is incorrect, and has lead to a lot of confusion where people are trying to build C++ mods against main branch, while having stable version of ue4ss installed, which don't work.

Adds proper caution tags to the top of the readme depending on the version being viewed

Also edit readmes for non-dev versions that reflects the correct way to clone the repository

**How Has This Been Tested?**

1. Run following commands in order:
`cargo binstall --no-confirm mdbook-mermaid`
`mdbook-mermaid install docs-export/`
`cargo binstall --no-confirm mdbook-alerts`
`python docs-export/export.py`
`python docs-repo/build.py`
2. Open `docs-repo/docs/` and then open index.html and view readme

Release readme changes:
image
image

Dev readme:
image
No change to the build instructions

Any release tag version (e.g. v3.0.0):
image
image

### DIFF
--- a/docs-export/export.py
+++ b/docs-export/export.py
@@ -46,18 +46,48 @@ def export_version(ref, name):
     subprocess.check_output(('tar', 'xvf', '-', '-C', src_dir), stdin=ps.stdout)
     ps.wait()
 
-    # copy README.md to <version>/src/
-    # rewrite absolute URLs to relative mdBook URLs
-    copy_transform('README.md', os.path.join(src_dir, 'README.md'), lambda content: re.sub(r'\(https://docs.ue4ss.com/dev/([^)#]+)\.html(#[^)]*)?\)', r'(\1.md\2)', content))
+    readme_spec = f'{ref}:README.md'
+    readme_out_path = os.path.join(src_dir, 'README.md')
 
-    # insert banner
-    if name != 'release':
-        readme_path = os.path.join(src_dir, 'README.md')
-        with open(readme_path, 'r') as file:
-            lines = file.readlines()
-        lines.insert(2, '\n> WARNING: This is the dev version of the UE4SS docs. The API and features are subject to change at any time. If you are developing mods for UE4SS, you should reference the [latest release](../release) instead.\n\n')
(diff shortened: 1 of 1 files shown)


igromanru:
> Currently, ue4ss home page and release tag (i.e. v3.0.1) shows the instructions for building ue4ss against main branch, which is incorrect

Maybe it's not the right place to discuss, but in my opinion people should use and make mods for the experimental version.
In experimental version a lot of bugs were fixed + all new game releases which use 5.3 and up require it anyway.
I would even like to suggest to release a new stable version ASAP or make the experimental version easier to find.
The stable version doesn't even work for some games like Hogwarts Legacy (4.27) anymore. I can't remember the reason but the "stable" version is simply old and a lot happened on all ends (Pseudo, patternsleuth etc.) 
Also as mod developer I find xmake easier to use and I'm glad that I don't have to install and deal with cmake. It might change back to cmake later, but currently xmake is alright as long there are no problems.

Buckminsterfullerene02:
> > Currently, ue4ss home page and release tag (i.e. v3.0.1) shows the instructions for building ue4ss against main branch, which is incorrect
> 
> Maybe it's not the right place to discuss, but in my opinion people should use and make mods for the experimental version. In experimental version a lot of bugs were fixed + all new game releases which use 5.3 and up require it anyway. I would even like to suggest to release a new stable version ASAP or make the experimental version easier to find. The stable version doesn't even work for some games like Hogwarts Legacy (4.27) anymore. I can't remember the reason but the "stable" version is simply old and a lot happened on all ends (Pseudo, patternsleuth etc.) Also as mod developer I find xmake easier to use and I'm glad that I don't have to install and deal with cmake. It might change back to cmake later, but currently xmake is alright as long there are no problems.

You are very correct, and I think that perhaps this PR should only be merged after the next major release. Because this is still a bug with the docs site, but I think that for now having the "wrong" readme at the front is fine.

PR #718, opened 2024-12-13
Include sorting

UE4SS:
**Description**

This adds include sorting and ordering to UE4SS, and its dependencies except Unreal, patternsleuth, patternsleuth bind and all third-party dependencies.

Before: https://github.com/UE4SS-RE/RE-UE4SS/blob/8f077f6bf92aed0ccb359f916d981c3a63b8cefc/UE4SS/src/UE4SSProgram.cpp
After: https://github.com/UE4SS-RE/RE-UE4SS/blob/968f6fd/UE4SS/src/UE4SSProgram.cpp

**How Has This Been Tested?**

Successfully built with msvc-wine.
Someone should test this on Windows natively.


Buckminsterfullerene02:
Builds fine for me on windows and on msvc-wine

localcc:
Looks good, I would squish the comments into two though, first one being adding clang-format and running it with all the build fixes, and the second one being git-blame-revs changes

narknon:
Can you do the existing Clang format first in one PR and then the include things separate?

UE4SS:
> Can you do the existing Clang format first in one PR and then the include things separate?

Okay, please merge #719 so that I can redo this PR.

UE4SS:
This should be ready now.

Buckminsterfullerene02 (approved):

narknon:
Does it have to Alpha sort all groups? Also can imagine some scenarios where Unreal includes should be the first group.

UE4SS:
> Does it have to Alpha sort all groups? Also can imagine some scenarios where Unreal includes should be the first group.

1. Why shouldn't it be alpha sorted ?
2. What scenarios ? The entire project builds as it is.

UE4SS:
> > Does it have to Alpha sort all groups?

I just checked, and as far as I could tell it's not possible to have groups without sorting the contents of each group.

narknon:
I don't see any benefit at all to alpha sorting, but there are plenty of potential scenarios where it could cause issues. For example, CoreMinimal should almost always be first include for the Unreal group and there are files that start with A. 

A lot of our types, defines and macros come from the unreal submodule and if they're included later there could be issues with defines not being applied early enough. 

Sometimes there are files that need to be included in a certain order, particularly around windows includes.  Generally they begin with Allow/Hide/Show.  Sometimes one or the other needs to be included at the very bottom of a file. This is just one scenario I know of, could be tons of other things we are not aware of.

PR #721, opened 2024-12-19
Better error messages

UE4SS:
**Description**

This PR makes various improvements to error messages.

**Screenshots**

Before
image

After
image

### DIFF
--- a/UE4SS/include/LuaType/LuaUObject.hpp
+++ b/UE4SS/include/LuaType/LuaUObject.hpp
@@ -251,6 +251,42 @@ namespace RC::LuaType
             return "UnknownOperation";
         }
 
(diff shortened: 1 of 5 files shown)


narknon:
Not sure if relevant to this PR, but it would be nice if when you tried to get data from a table using an incorrect name the log printed out all the actual names in the table.  E.g., an out param for a struct that you then want to get a particular value from.

UE4SS:
> Not sure if relevant to this PR, but it would be nice if when you tried to get data from a table using an incorrect name the log printed out all the actual names in the table. E.g., an out param for a struct that you then want to get a particular value from.

I'm not prepared to make such a change in this PR.
I'm also not sure how it would be implemented because the table used for out params is provided by the script, and it's just a table.

narknon (approved):

PR #727, opened 2024-12-27
chore: include mod names in logs of built-in Lua mods

massclown:
**Description**

Add `[ModName] ` to the log messages of all built-in mods that have log messages, so it is 100% clear which mod does each error message come from. Also added a few missing line endings to log messages here and there.

**Type of change**

- [Y] Other... Please describe: minor refactoring

**How Has This Been Tested?**

I enabled all built-in mods and ran them.

**Checklist**

- [Y] I have added the necessary description of this PR to the changelog, and I have followed the same format as other entries.


UE4SS (approved):
This looks fine.
Note however that these changes will need to be reverted in the future if we add this functionality directly to the `print` function using the `ModRef` data.
This PR is small so I don't mind reverting manually in the future.

PR #738, opened 2024-12-30
Fix experimental-latest release making

Fancy2209:
**Description**

Fixes the tag 

**Type of change**

- [X] Other... Please describe: Ci Fix

**How Has This Been Tested?**

I haven't yet, Github Actions are very annoying, but the glob now matches the file name so it should just work

### DIFF
--- a/.github/workflows/experimental.yml
+++ b/.github/workflows/experimental.yml
@@ -91,7 +91,7 @@ jobs:
           tag_name: experimental-latest
           body_path: release/release_notes.md
           files: |
-            release/UE4SS_v*.zip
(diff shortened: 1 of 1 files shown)


narknon:
Thank you! This actually seems to be caused by an issue with the release.py script related to it just grabbing the newest tag, so I'm testing a change there.

narknon:
The other fix worked. I'm debating if there's any other reason we may want to drop the "v" as it isn't really necessary.  I guess having it did alert us to the tag issue sooner though.

narknon:
Closing for now because the bug was fixed in another location, may reopen if we decide to drop the v*.

PR #739, opened 2024-12-31
Fix blank spaces in live view scroll

narknon:


UE4SS:
I will not review this PR unless you explain what this is, and why it fixes whatever problem this is referring to.

narknon:
It fixes the bug introduced with the PR that reduced live view lag where there would be blank lines and spaces when scrolling up and down.

UE4SS:
Okay since you only gave half an answer, I'll have to guess based on what I've seen, please let me know if I got it right.

Previously, when using objects are filtered without searching, it wasn't taking into account the filtered objects when configuring the clipper, as a result, empty elements would be rendered for those filtered objects.
This PR introduces separate logic for when objects are filtered without searching.

Correct ?

narknon:
The bug was when there was no filter. I think the blank line bug was due to passing in the guobjectarray size which didn't take into account objects that become unreachable. This makes a new vector without unreachable objects to get an accurate size to iterate for the scroll list.

narknon:
I.e., object becomes unreachable, index still exists and ImGui still thinks it needs to render a line, but the object is unreachable so we don't pass any info about it in.

UE4SS:
> I.e., object becomes unreachable, index still exists and ImGui still thinks it needs to render a line, but the object is unreachable so we don't pass any info about it in.

Checking if it's reachable should be done regardless I think, so we don't really need both paths here.
I think it's appropriate to remove the original code completely.

narknon:
Well any other path that reaches it uses foreachuobject which does check if it's unreachable, though actually, yeah, while typing this out I'm realizing it may still pass in the wrong size/indices, I'm not sure

narknon:
Confirming that the issue does not seem to exist with a filter applied.

PR #747, opened 2025-01-07
Default value of 'bUseUObjectArrayCache' is now false in non-zDev

### LINKED ISSUE #315: Performance issues
What if this absolutely destroys the performance of the game? The palworld community uses this for its modding, i have asked over on their reddit a couple days ago but got no response. I lose half my frames, get stutters, the game gets stuck for a dozen of seconds when it has to do any type of loading, like when starting, loading world (which sometimes crashes the game), and going back to the titel, and if i idle on titel. This happens with or without mods.

UE4SS:
**Description**

The `bUseUObjectArrayCache` option doesn't work as intended, and fixing it seems unlikely so this PR disables it for non-zDev releases by default.
It still needs to be enabled for zDev releases because the live view search feature relies on it.

Fixes #315

**Type of change**

- [ ] Other... Please describe: Packaging update

**How Has This Been Tested?**

Locally ran the release.py srcipt and verified that the value of the option is correct.

**Checklist**


narknon:
I wouldn't say this Fixes #315, it's more of a workaround for now.

UE4SS:
> I wouldn't say this Fixes #315, it's more of a workaround for now.

Okay.

igromanru:
Oh, that's how it's done. Makes sense, default value is always `true` and only for release it overwrites with `false`.

PR #749, opened 2025-01-08
InputSource, replaces #558

UE4SS:
This PR is rebased onto main.
The old one is a disaster of a PR, this one contains only the InputSource commits and is rebased onto main.
Please see #558 for details.

This PR needs testing.
Keybinds seem to work fine for me on Linux using Proton, tested with Palworld.
Someone needs to test on Windows natively.

### DIFF
--- a/UE4SS/include/UE4SSProgram.hpp
+++ b/UE4SS/include/UE4SSProgram.hpp
@@ -11,7 +11,10 @@
 #include <DynamicOutput/DynamicOutput.hpp>
 #include <GUI/GUI.hpp>
 #include <GUI/GUITab.hpp>
+#ifdef HAS_INPUT
 #include <Input/Handler.hpp>
+#endif
+
 #include <LuaLibrary.hpp>
 #include <MProgram.hpp>
 #include <Mod/CppMod.hpp>
@@ -98,7 +101,9 @@ namespace RC
         bool m_is_program_started;
 
       protected:
-        Input::Handler m_input_handler{L"ConsoleWindowClass", L"UnrealWindow"};
+#ifdef HAS_INPUT
+        Input::Handler m_input_handler;
+#endif
         std::jthread m_event_loop;
 
       public:
@@ -221,6 +226,12 @@ namespace RC
         RC_UE4SS_API auto generate_uht_compatible_headers() -> void;
         RC_UE4SS_API auto generate_cxx_headers(const std::filesystem::path& output_dir) -> void;
         RC_UE4SS_API auto generate_lua_types(const std::filesystem::path& output_dir) -> void;
+#ifdef HAS_INPUT
+        auto get_input_handler() -> Input::Handler&
+        {
+            return m_input_handler;
+        }
+#endif
         auto get_debugging_ui() -> GUI::DebuggingGUI&
         {
             return m_debugging_gui;
@@ -244,6 +255,7 @@ namespace RC
         }
 
       public:
+#ifdef HAS_INPUT
         // API pass-through for use outside the private scope of UE4SSProgram
         RC_UE4SS_API auto register_keydown_event(Input::Key, const Input::EventCallbackCallable&, uint8_t custom_data = 0, void* custom_data2 = nullptr) -> void;
         RC_UE4SS_API auto register_keydown_event(Input::Key,
@@ -253,6 +265,7 @@ namespace RC
                                                  void* custom_data2 = nullptr) -> void;
         RC_UE4SS_API auto is_keydown_event_registered(Input::Key) -> bool;
         RC_UE4SS_API auto is_keydown_event_registered(Input::Key, const Input::Handler::ModifierKeyArray&) -> bool;
+#endif
 
       private:
         static auto install_cpp_mods() -> void;
--- a/UE4SS/src/Mod/CppUserModBase.cpp
+++ b/UE4SS/src/Mod/CppUserModBase.cpp
@@ -3,6 +3,8 @@
 #include <UE4SSProgram.hpp>
 #include <String/StringType.hpp>
 
+#include <vector>
+
 namespace RC
 {
     CppUserModBase::CppUserModBase()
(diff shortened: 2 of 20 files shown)


narknon:
How crazy would it be to make the register_keydown_event and other functions/lambdas just compile into nothing rather than wrapping every use of them in ifdefs?

UE4SS:
> How crazy would it be to make the register_keydown_event and other functions/lambdas just compile into nothing rather than wrapping every use of them in ifdefs?

I agree.
I think it's better to wrap the contents of the functions with #ifdef instead so that it's empty when not supported.

UE4SS:
> > How crazy would it be to make the register_keydown_event and other functions/lambdas just compile into nothing rather than wrapping every use of them in ifdefs?
> 
> I agree. I think it's better to wrap the contents of the functions with #ifdef instead so that it's empty when not supported.

Although this could be tricky with functions that use a type that is also #ifdef'd out, like `get_input_handler`.
You'd have to dummy that type to make it compile, and then it also returns a reference which means there needs to be an `#else` that returns a static version of the dummy type.
It makes it more complicated to implement but removes all burden from users of the API which I think is better.

narknon:
Can you just ifdef the members?

UE4SS:
> Can you just ifdef the members?

I don't see how that would help here.
If the members are ifdef'd out, anything that uses those members won't compile.

narknon:
Won't the only things using the members be the ifdef'd functions? Idk, I haven't looked closely at it.

UE4SS:
> Won't the only things using the members be the ifdef'd functions? Idk, I haven't looked closely at it.

Members and functions are currently fully idef'd out.
As a result, users of those members and functions also need an ifdef around them or they won't compile.
This would be inconvenient for mods that register keybinds for example.
They shouldn't have to ifdef their own calls to the registration function.
If we remove `get_input_handler`, I think the registration functions are the only place where this matters as everything else is internal stuff.

UE4SS:
The discussion above has been resolved, the current status is now:
1. Waiting for someone to test and confirm that nothing broke on native Windows.
2. Waiting for @Yangff to possibly have a quick look.

igromanru:
I tested the latest build from here with my mods, mainly with my StackManage mod, which utilize multiple `RegisterKeyBind` with same primary key and different modifiers.  
Couldn't find any issues.

UE4SS/include/UE4SSProgram.hpp
```diff
@@ -12,6 +12,7 @@
 #include <GUI/GUI.hpp>
 #include <GUI/GUITab.hpp>
 #include <Input/Handler.hpp>
+
```
UE4SS, line 15: Note to self: Remove empty line before merging.

UE4SS/src/Mod/LuaMod.cpp
```diff
@@ -11,7 +11,9 @@
 #include <ExceptionHandling.hpp>
 #include <Helpers/Format.hpp>
 #include <Helpers/String.hpp>
+
 #include <Input/Handler.hpp>
+
```
UE4SS, line 16: Note to self: Remove empty lines before merging.

UE4SS/src/Mod/CppUserModBase.cpp
```diff
@@ -3,6 +3,8 @@
 #include <UE4SSProgram.hpp>
 #include <String/StringType.hpp>
 
+#include <vector>
```
UE4SS, line 6: Note to self: Move to the top of the file to be inline with our general include order.

deps/first/Input/include/Input/Platform/GLFW3InputSource.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+#include <Input/Platform/QueueInputSource.hpp>
+#include <vector>
```
UE4SS, line 4: Note to self: Move to the top of the file to be inline with our general include order.

deps/first/Input/include/Input/Platform/QueueInputSource.hpp
```diff
@@ -0,0 +1,61 @@
+#pragma once
+
+#include <Input/PlatformInputSource.hpp>
+#include <Input/KeyDef.hpp>
+
```
UE4SS, line 5: Note to self: Remove empty line

deps/first/Input/include/Input/Platform/Win32AsyncInputSource.hpp
```diff
@@ -0,0 +1,86 @@
+#pragma once
+
+#include <vector>
+#include <unordered_map>
+#include <Input/KeyDef.hpp>
+#include <Input/Common.hpp>
+
```
UE4SS, line 7: Note to self: Remove empty line

deps/first/Input/include/Input/Platform/Win32AsyncInputSource.hpp
```diff
@@ -0,0 +1,86 @@
+#pragma once
+
+#include <vector>
+#include <unordered_map>
```
UE4SS, line 4: Note to self: Add empty line

deps/first/Input/include/Input/PlatformInputSource.hpp
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include <Input/Handler.hpp>
+
+#include <optional>
+#include <vector>
```
UE4SS, line 6: Note to self: Move to the top of the file to be inline with our general include order.

deps/first/Input/src/KeyDef.cpp
```diff
@@ -1,4 +1,5 @@
 #include <Input/KeyDef.hpp>
+#include <cstdint>
```
UE4SS, line 2: Note to self: Move to the top of the file to be inline with our general include order.

deps/first/Input/src/Platform/QueueInputSource.cpp
```diff
@@ -0,0 +1,46 @@
+#include <Input/PlatformInputSource.hpp>
+#include <Input/Platform/QueueInputSource.hpp>
+
```
UE4SS, line 3: Note to self: Remove empty line

narknon:
If you don't mind, please manually squash the last two commits then do a normal merge. Not a huge deal though.

PR #751, opened 2025-01-10
Fixed indexing nil in BPModLoaderMod

UE4SS:
**Description**

Sometimes, for some unknown reason, the `World` argument to `LoadMod` can be nil.
This must be checked for or Lua will throw an error.

**Type of change**

**How Has This Been Tested?**

It has not beyond making sure the syntax is valid.

**Checklist**

### DIFF
--- a/assets/Changelog.md
+++ b/assets/Changelog.md
@@ -196,6 +196,8 @@ Fixed `RegisterProcessConsoleExecPostHook`. ([UE4SS #631](https://github.com/UE4
 
 Fixed `FindFirstOf` return type annotation in `Types.lua` to signal that the return value will never be nil. ([UE4SS #652](https://github.com/UE4SS-RE/RE-UE4SS/issues/652)) 
 
+Fixed indexing nil in BPModLoaderMod ([UE4SS #751](https://github.com/UE4SS-RE/RE-UE4SS/issues/751))
+
 ### C++ API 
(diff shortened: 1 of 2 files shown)


igromanru:
I think it's not wrong to check for nil there, but has it actually happen that it was `nil`?
Theoretically it shouldn't happen with UEHelpers v3, since we have a fallback.

UE4SS:
> I think it's not wrong to check for nil there and I approve the change, but has it actually happen that it was `nil`? Theoretically it shouldn't happen with UEHelpers v3, since we have a fallback in GetWorld() now.
> 
> EDIT: I'm asking because if World was nil at some point there, it might be an indicator that something else it wrong. For example: How good does the scope of function parameter `World` gets transferred to async callback of `ExecuteInGameThread`?

Yes, check discord: https://discord.com/channels/1148976194050605066/1148976979048149044/1327317826448527400
Though I don't know what version of UEHelpers they are using.

igromanru:
We have discussed it in Discord that the underline problem is most likely that function variable `World` goes out of scope in async callback of `ExecuteInGameThread`. It might work if the callback gets called immediately, but since we have no control when `ExecuteInGameThread` gets it turn in the queue, Lua's garbage collector probably removes `World` at some point.  
I want to improve scripts behavior and make own PR for it.  
You can either close the RP or just merge as it is.

PR #762, opened 2025-01-29
fix(Lua): Key::NUM_ZERO being mapped incorrectly in Lua

UE4SS:
**Type of change**

**Checklist**


igromanru:
Looks right.
Sure that it's the last place where virtual keys are defined?

UE4SS:
> Looks right. Sure that it's the last place where virtual keys are defined?

I searched and couldn't find any other usages.

PR #768, opened 2025-01-31
Option for rendering GUI in the game thread

### LINKED ISSUE #764: [BUG - Release] Tokyo Xtreme Racer (UE5.4): Live View frequently crashes
**Branch or Release**
experimental-latest, at the time of submitting this issue that's the commit 5015114a132de4f410b26b108afd8bee3281bc24.

**Game and Engine Version**
Tokyo Xtreme Racer, running on UE 5.4.4.0. GUI Console uses D3D11 to render.

**Describe the bug**
When previewing actor data in the game via Live View, the window crashes frequently. The game itself remains operational unless the crash window is dismissed, but the GUI Console itself stops working.

**Mods directory**
Only default mods are present, and all but `ConsoleEnablerMod` and `Keybinds` are disabled.

**To Reproduce**
1. Launch the game.
2. Open the Live View.
3. Wait, the game will eventually crash.

**Expected behavior**
UE4SS doesn't crash and it lets me preview actors.

At the time of the crash I was not even focused on the console window, and was not scrolling/editing anything:
Image

Log and crashdump:
[txr2025-live-preview-crash.zip](https://github.com/user-attachments/files/18602479/txr2025-live-preview-crash.zip)

**Desktop:**
 - OS: Windows 10 22H2

**Additional context**
-

UE4SS:
**Description**

This PR adds an option to render the GUI in the game thread.
The default is still to render in its own GUI thread.
To turn on game thread rendering, set `RenderMode` in UE4SS-settings.ini to `GameViewportClientTick` in the `[Debug]` section.

Fixes #764 

**Type of change**

**How Has This Been Tested?**

With RenderMode set to `GameViewportClientTick`, start the game, make sure the GUI opens, and use CTRL + O to close it and again to open it.
Repeat with `RenderMode` set to `ExternalThread`.

**Checklist**

**Additional context**

This needs to be tested for a bunch of games, and ideally by more than one person.

### DIFF
--- a/UE4SS/include/GUI/GUI.hpp
+++ b/UE4SS/include/GUI/GUI.hpp
@@ -22,6 +22,12 @@ namespace RC::GUI
         GLFW3_OpenGL3,
     };
 
+    enum class RenderMode
+    {
+        ExternalThread,
+        GameViewportClientTick,
+    };
+
     enum class OSBackend
     {
         Windows,
@@ -211,7 +217,7 @@ namespace RC::GUI
         std::unique_ptr<OSBackendBase> m_os_backend{};
         Console m_console{};
         LiveView m_live_view{};
-        std::stop_token m_thread_stop_token{};
+        std::stop_token* m_thread_stop_token{};
         bool m_is_open{};
         bool m_exit_requested{};
         std::vector<std::shared_ptr<GUITab>> m_tabs;
@@ -230,12 +236,18 @@ namespace RC::GUI
         ~DebuggingGUI();
 
       public:
-        auto is_valid() -> bool;
-        auto is_open() -> bool
+        [[nodiscard]] auto is_valid() const -> bool;
+        [[nodiscard]] auto is_open() const -> bool
         {
             return m_is_open;
-        };
-        auto setup(std::stop_token&& token) -> void /* override*/;
+        }
+        auto set_open(bool new_open) -> void;
+        [[nodiscard]] auto exit_requested() const -> bool
+        {
+            return m_exit_requested;
+        }
+        auto request_exit() -> void;
+        auto setup(std::stop_token* token) -> void /* override*/;
         auto get_console() -> Console&
         {
             return m_console;
@@ -247,10 +259,11 @@ namespace RC::GUI
         auto set_gfx_backend(GfxBackend) -> void;
         auto add_tab(std::shared_ptr<GUITab> tab) -> void;
         auto remove_tab(std::shared_ptr<GUITab> tab) -> void;
+        auto uninitialize() -> void;
+        auto main_loop_internal() -> void;
 
       private:
         auto on_update() -> void;
-        auto main_loop_internal() -> void;
 
       private:
         // TODO: Move ImGui data saves to their own object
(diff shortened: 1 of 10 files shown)


CookiePLMonster:
Tested **Tokyo Xtreme Racer** running on UE5.4, with Live View on-screen:

`RenderMode = ExternalThread` - crashes easily, same as in #764, **as expected**.
`RenderMode = GameViewportClientTick` - I can't reproduce the crash after Alt+Tabbing frantically and scrolling through the list. Additionally, Alt+Tab during the loading screen appears to work fine now, instead of crashing the game.

With `GameViewportClientTick`, the game pauses briefly when filtering the search results, but again that is expected behaviour.

igromanru:
How do I know if GameViewportClient::Tick hook works? I see nothing in console output about it.  
However, I tested it in Abiotic Factor, where poking around and keeping the Live View open crashes the game randomly after some minutes.  
After setting `RenderMode = GameViewportClientTick` I was jumping around between different objects then let it idle for about 10min and it never crashed. So I guess it works.

CookiePLMonster:
> How do I know if GameViewportClient::Tick hook works? I see nothing in console output about it. However, I tested it in Abiotic Factor, where poking around and keeping the Live View open crashes the game randomly after some minutes. After setting `RenderMode = GameViewportClientTick` I was jumping around between different objects then let it idle for about 10min and it never crashed. So I guess it works.

Try searching in Live View. With enough objects, the game will likely pause for a bit as the results are filtered. This only happens when the external thread is **not** used.

UE4SS:
> How do I know if GameViewportClient::Tick hook works? I see nothing in console output about it.

I don't think we generally report what hooks are enabled.
The best way to tell which mode is being used is to pay attention to when the GUI opens.
If it opens immediately, it's ExternalThread, it it doesn't open until after the games loading screen has appeared then it's GameViewportClientTick.

igromanru:
> Try searching in Live View. With enough objects, the game will likely pause for a bit as the results are filtered. This only happens when the external thread is **not** used.

I've no control about how many objects are loaded. It competently depends on the game. And in my game, even when I'm in the center of my base, which has a whole bunch of objects around it, I don't have freezes when filtering.
Therefore, I would like to have a console output which hooks are activated and actually working.  
Because I can only guess right now since it doesn't crash anymore.

UE4SS:
> > Try searching in Live View. With enough objects, the game will likely pause for a bit as the results are filtered. This only happens when the external thread is **not** used.
> 
> I've no control about how many objects are loaded. It competently depends on the game. And in my game, even when I'm in the center of my base, which has a whole bunch of objects around it, I don't have freezes when filtering. Therefore, I would like to have a console output which hooks are activated and actually working. Because I can only guess right now since it doesn't crash anymore.
> 
> > If it opens immediately, it's ExternalThread, it it doesn't open until after the games loading screen has appeared then it's GameViewportClientTick.
> 
> I disabled the into movie, therefore it's instantly anyway. But yeah, looks like the GUI appears once the game is loaded.

I've added the render mode to the window title, next to dx11/opengl.

narknon (approved):

PR #769, opened 2025-01-31
Added C++/Lua hooks for AActor::Endplay

UE4SS:
**Description**

This PR adds hooks accessible via both Lua and C++ for `AActor::EndPlay`, in the same way that we have `AActor::BeginPlay`.

**Type of change**

**How Has This Been Tested?**

Make a Lua mod, register an EndPlay hook that prints the object name.
Start a game, load some save files, and quit.
Open UE4SS.log and make sure the output is correct.

**Checklist**

**Additional context**

This may not be merged until #768 is merged.
If #768 doesn't get merged, this PR must be rebased onto main.


UE4SS:
Note that CI fails only because the branch is based on a commit that doesn't have the latest Unreal module commit.
It will pass when #768 is merged, and/or this branch is rebased onto main.

CookiePLMonster:
Confirmed working:
image

PR #788, opened 2025-02-11
Add FText function description to Types.lua and rewrite description in Docs

igromanru:
**Description**
I've noticed that the example of the use of `FText` object in `FText` function documentation was wrong, and the description overall is weirdly worded ("FText representation").  
I fixed the example code and improved the description to make it easier to understand what the function does.  
At the same time `FText` function was missing in the `Types.lua`.

### DIFF
--- a/docs/lua-api/global-functions/ftext.md
+++ b/docs/lua-api/global-functions/ftext.md
@@ -1,27 +1,34 @@
 # FText
 
-The `FText` function is used to get an `FText` representation of a `string`.
+The `FText` function is used to create a `FText` object from a `string`.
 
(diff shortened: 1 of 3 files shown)


docs/lua-api/global-functions/ftext.md
```diff
@@ -1,27 +1,34 @@
 # FText
 
-The `FText` function is used to get an `FText` representation of a `string`.
+The `FText` function is used to create a `FText` object from a `string`.
```
narknon, line 3: Just a weird language thing, but this should be 'an'.  Although FText starts with a consonant, phonetically it starts with a vowel (ef-text).

igromanru: Fixed

UE4SS (approved):

PR #790, opened 2025-02-12
feat(docs): Added list of games that PS supports out of the box

UE4SS:
**Description**

This PR adds a list of games that patternsleuth supports out of the box.

**Type of change**

- [x] Other... Please describe: docs


UE4SS:
> The page should be called something other than "supported_games"

The file has now been renamed.

localcc:
having a compatibility page wouldn't necessarily be bad but it should be made in a better formatted way with a version of the game clearly listed and preferrably date tested as well, I would draw inspiration from something like protondb or compatibility lists of various emulators. Just having a big list of exe names isn't too helpful I think. Also the list should be less a "compatibility" list and more of a "well known working" one, as compatibility implies anything not in the list is incompatible

igromanru:
> having a compatibility page wouldn't necessarily be bad but it should be made in a better formatted way with a version of the game clearly listed and preferrably date tested as well, I would draw inspiration from something like protondb or compatibility lists of various emulators. Just having a big list of exe names isn't too helpful I think. Also the list should be less a "compatibility" list and more of a "well known working" one, as compatibility implies anything not in the list is incompatible

**This is my original feedback that I wrote to Martin in DM:**
If you generate the list automatically with an own script, I would like to suggest making a Markdown table out of it.
Like:
```
Steam ID | Game Name | Info
-------- | --------- | -----
```
Then you can just make two tables this way. "Games tested with all AOBs found by Patternsleuth" and "Games tested but with AOBs not found by Patternsleuth"
If you have no way to get the Steam ID, you can leave it blank for now, and it can be added manually later or the script can also be extended to search the game on steam and get the Steam ID automatically. Maybe even wrap the ID into a link to the SteamDB or the Steam page.  
It's pretty easy to get gane names and their IDs with the Steam API:
<http://api.steampowered.com/ISteamApps/GetAppList/v2/?key=STEAMKEY&format=json>
Be careful opening the link, it contains a huge amount of data.


I think the list is currently good enough, better than nothing. But overall it would be nice to add a column or something that can be edited later, when people actually test the game. Because PS AOB support doesn't guarantee UE4SS to work. There are also other elements like default Mods that has to be tested for compatibility.

docs/SUMMARY.md
```diff
@@ -125,5 +125,6 @@
 ## Misc
 
 - [Custom Game Configs](./custom-game-configs.md)
+- [Games with AOB support](./supported_games.md)
```
Buckminsterfullerene02, line 128:
```suggestion
- [Games with AOB support](./patternsleuth_supported_games.md)
```

UE4SS: Good catch, thanks.

Buckminsterfullerene02:
I think this is a useful page to have - the warnings are clear enough, and if anyone complains that they saw their game on the PS list but doesn't work OOTB, then they can simply be told to read the warnings. I think the benefits outweigh the problem.

Buckminsterfullerene02 (approved):

PR #794, opened 2025-02-17
Add HookEngineTick setting and EngineTick as GUI RenderMode

igromanru:
**Description**
Similar to #768 where `GameViewportClientTick` was added as a new `RenderMode` for the GUI Tool, I've added `EngineTick`.
Additionally, I had to add `HookEngineTick` setting, as it wasn't a thing yet.  
I put `EngineTick` everywhere in front of `GameViewportClientTick` because I think it's more important. GameViewportClient doesn't exist on servers and with Engine::Tick implementation the hook might be obsolete in the future.

**How it was tested**
After setting `RenderMode = EngineTick`, I attached the debugger to the game at the start and made sure that `Engine::Tick` is used instead of `GameViewportClient::Tick`.  
Then I played around in the Live View like before and it worked well. At least as good as it usually works.
Additionally, I tested it in Dedicated Server and unlike `GameViewportClientTick` it works there as well! 🥳
The game in question is Abiotic Factor.
AbioticFactor-Win64-Shipping_AhZ5iE1ids

### DIFF
--- a/assets/Changelog.md
+++ b/assets/Changelog.md
@@ -39,7 +39,6 @@ Added custom game configurations for Psychonauts 2 ([UE4SS #731](https://github.
 The GUI can now be rendered in the game thread if `RenderMode` in UE4SS-settings.ini is set to
 `EngineTick` or `GameViewportClientTick` ([UE4SS #768](https://github.com/UE4SS-RE/RE-UE4SS/pull/768), [UE4SS #794](https://github.com/UE4SS-RE/RE-UE4SS/pull/794)).
 
-Added new setting `HookEngineTick` ([UE4SS #794](https://github.com/UE4SS-RE/RE-UE4SS/pull/794)).
 
 ### Live View 
 Added search filter: `IncludeClassNames`. ([UE4SS #472](https://github.com/UE4SS-RE/RE-UE4SS/pull/472)) - Buckminsterfullerene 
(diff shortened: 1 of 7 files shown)


assets/Changelog.md
```diff
@@ -39,6 +39,10 @@ Added custom game configurations for Psychonauts 2 ([UE4SS #731](https://github.
 The GUI can now be rendered in the game thread if `RenderMode` in UE4SS-settings.ini is set to
 `GameViewportClientTick` ([UE4SS #768](https://github.com/UE4SS-RE/RE-UE4SS/pull/768)).
 
+Added `EngineTick` `RenderMode` for the GUI ([UE4SS #794](https://github.com/UE4SS-RE/RE-UE4SS/pull/794)).
+
+Added new setting `HookEngineTick` ([UE4SS #794](https://github.com/UE4SS-RE/RE-UE4SS/pull/794)).
```
UE4SS, line 44:
```suggestion
```
We don't have any entries in the changelog regarding ini changes outside the `Settings` section, so I think this changelog entry should be removed.

UE4SS (changes requested):

UE4SS (approved):
Approved.
Waiting a couple days to let other people have a chance to see this.
Note to self: Clean up the two changelog commits that I created before merging.

PR #801, opened 2025-02-20
Performance improvements for script hooks

UE4SS:
**Description**

This PR improved the performance of script hooks.
We were calling `GetFullName` every time `ProcessLocalScriptFunction` (4.22+) or `ProcessInternal` (<4.22) was called.
This is an expensive operation and should be avoided whenever possible.
This PR gets rid of this call in favor of comparing the FName of each outer of the function.

**Type of change**

- [ ] Other... Please describe: Performance improvement

**How Has This Been Tested?**

I did a very basic script hook with `RegisterHook` in Lua.
Please do some testing to make sure this PR doesn't break stuff.

**Checklist**

### DIFF
--- a/UE4SS/include/Mod/LuaMod.hpp
+++ b/UE4SS/include/Mod/LuaMod.hpp
@@ -15,6 +15,8 @@
 
 #include <String/StringType.hpp>
 
+#include <Unreal/NameTypes.hpp>
+
 namespace RC
 {
     class UE4SSProgram;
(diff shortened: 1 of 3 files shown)


igromanru:
No idea if it's a proper way to test it. But I did following:
I used jsbProfiler to register 50 times the same hook. I waited until the game is fully loaded, then used "Mods Reload" button to reload all mods and test the performance of it after the game was already fully loaded.
The results from jsbProfiler aren't very good, but here it is.
**Code**
```lua
jsb.simpleBenchMulti("RegisterHook -> UpdateChangeableState", 50, RegisterHook, "/Game/Blueprints/Items/FakeItems/CookingProxy_BP.CookingProxy_BP_C:UpdateChangeableState", function(Context, CookingState, CookedProgress)
    local context = Context:get() ---@type ACookingProxy_BP_C
    local cookingState = CookingState:get() ---@type EFoodCookStates
    local cookedProgress = CookedProgress:get() ---@type float
    
    LogDebug("----- [UpdateChangeableState] called -----")
    LogDebug("CookingState:", cookingState)
    LogDebug("CookedProgress:", cookedProgress)
    LogDebug("State:", context.State)
    LogDebug("Cookware:", context.Cookware)
    LogDebug("IsCookingSoup:", context.IsCookingSoup)
    LogDebug("------------------------------")
end)
```

**zDEV-UE4SS_v3.0.1-314-gdd2da52 Result**
```
Thu Feb 20 12:35:15 2025: New test session

Thu Feb 20 12:35:15 2025: Func RegisterHook -> UpdateChangeableState took: 0.001 and 2e-05 average per call
```
**MSVC-Game__Shipping__Win64 from here Result**
```
Thu Feb 20 12:34:30 2025: New test session

Thu Feb 20 12:34:30 2025: Func RegisterHook -> UpdateChangeableState took: 0 and 0 average per call
```

**My conclusion**
I think since usually hooks shouldn't be registered and unregistered in a loop or each frame, it doesn't rally matter.  
But overall, it's nice to optimize code.

UE4SS:
I have no idea how to use jsbProfiler, so I don't even know if that code is profiling the correct thing.
This PR mostly affects the hook execution, not the registration, although the registration is also affected.
Your test code kind of seems like it might be profiling the call to RegisterHook which isn't the important part.

igromanru:
Of course it tests only hook registration.
You can't test the hook execution speed on Lua side.
It has to be done inside UE4SS.

PR #752, opened 2025-01-11
BPModLoaderMod: Fix problem with out-of-scope variables in async call and ensure that LoadMods is always called in a game thread

igromanru:
### Description
- Removed `ExecuteInGameThread` call from the `LoadMod` function.
- Ensured that `LoadMods` is always called in a game thread instead.
- Added `error` messages in case the `World` is **nil** in either the  `LoadMods` or `LoadMod` function.

According to the current mod logic with the new UEHelpers v3, the parameter `World` shouldn't be `nil` in either `LoadMods` or `LoadMod`. It can be an invalid UObject, but never `nil`. Therefore, if it should happen, we should assume that it is an error in  BPModLoaderMod's code.

### Quality control
Tested in Palworld v0.4.12 with [Mod Config Menu (UI)](https://www.nexusmods.com/palworld/mods/577) and [Basic MiniMap](https://www.nexusmods.com/palworld/mods/146) mods.
Here is a snippet from the log file without timestamps for better visibility:
```
Starting Lua mod 'BPModLoaderMod'
[Lua] [BPModLoaderMod] Mods/BPModLoaderMod/load_order.txt not present or no matching mods, loading all BP mods in random order.
[Lua] [BPModLoaderMod] DekBasicMinimap_P == table: 000001D4366428B0
[Lua] [BPModLoaderMod]     AssetName == ModActor_C
[Lua] [BPModLoaderMod]     AssetNameAsFName == FNameUserdata: 000001D436575B88
[Lua] [BPModLoaderMod]     Name == DekBasicMinimap_P
[Lua] [BPModLoaderMod]     AssetPath == /Game/Mods/DekBasicMinimap_P/ModActor
[Lua] [BPModLoaderMod] DekModConfigMenu_P == table: 000001D436643230
[Lua] [BPModLoaderMod]     AssetName == ModActor_C
[Lua] [BPModLoaderMod]     AssetNameAsFName == FNameUserdata: 000001D436575B88
[Lua] [BPModLoaderMod]     Name == DekModConfigMenu_P
[Lua] [BPModLoaderMod]     AssetPath == /Game/Mods/DekModConfigMenu_P/ModActor
Mod 'jsbLuaProfilerMod' disabled in mods.txt.
Starting Lua mod 'Keybinds'
Starting mods (from enabled.txt, no defined load order)...
Mod 'DekModConfigMenu' has enabled.txt, starting mod.
Event loop start
[Lua] [BPModLoaderMod] Loading mod: DekBasicMinimap_P
[...]

PR #808, opened 2025-02-23
BPModLoaderMod: Replace error with a warning message in the LoadMods function

igromanru:
**Description**
In #752 I added an error message for the case that the World parameter passed to functions `LoadMod` and `LoadMods` is `nil`, which technically should never happen.  
It looks like in `LoadMods` it's only partially true.  
At the end of the script, the mod tries to execute `LoadMods` as soon as possible, by finding any AActor and using its UWorld.
```lua
    local ExistingActor = FindFirstOf("Actor")
    if ExistingActor:IsValid() then
        LoadMods(ExistingActor:GetWorld())
    end
end)
```
While I'm unsure if it would be better or easier just to get the current World with `UEHelpers.GetWorld()`, I have seen that at the time of loading and running the mod in some games `ExistingActor:GetWorld()` returns `nil`.  
I have no idea how it could be possible that `ExistingActor:GetWorld()` returns a `nil` and not an invalid object, but in the game [Avowed](https://store.steampowered.com/app/2457220/Avowed/) it seems to happen.  

My first idea which is now implemented in this PR is to add a `not World:IsValid()` check to `LoadMods` and replace the error with a simple warning message, since later it works fine when a map gets loaded and `LoadMapPostHook` is triggered.

But now if I think about it, we might want to discuss, if we should just use `UEHelpers.GetWorld()` and overall how it can be that `ExistingActor:GetWorld()` return a `nil`.

### DIFF
--- a/assets/Mods/BPModLoaderMod/Scripts/main.lua
+++ b/assets/Mods/BPModLoaderMod/Scripts/main.lua
@@ -22,7 +22,7 @@ DefaultModConfig.AssetNameAsFName = UEHelpers.FindOrAddFName("ModActor_C")
 
 -- Checks if the beginning of a string contains a certain pattern.
 local function StartsWith(String, StringToCompare)
-    return string.sub(String,1,string.len(StringToCompare))==StringToCompare
+    return string.sub(String, 1, string.len(StringToCompare)) == StringToCompare
 end
 
 local function FileExists(file)
@@ -31,20 +31,20 @@ local function FileExists(file)
     return f ~= nil
 end
 
--- Reads lines from the specified file and returns a table of lines read. 
+-- Reads lines from the specified file and returns a table of lines read.
 -- Second argument takes a string that can be used to exclude lines starting with that string. Default ;
 local function LinesFrom(file, ignoreLinesStartingWith)
     if not FileExists(file) then return {} end
 
-    if ignoreLinesStartingWith == nil then 
+    if ignoreLinesStartingWith == nil then
         ignoreLinesStartingWith = ";"
     end
 
     local lines = {}
-    for line in io.lines(file) do 
-      if not StartsWith(line, ignoreLinesStartingWith) then
-        lines[#lines + 1] = line
-      end
+    for line in io.lines(file) do
+        if not StartsWith(line, ignoreLinesStartingWith) then
+            lines[#lines + 1] = line
+        end
     end
     return lines
 end
@@ -71,7 +71,7 @@ local function LoadModOrder()
     end
 
     if entriesAdded <= 0 then
-        Log(string.format("Mods/BPModLoaderMod/load_order.txt not present or no matching mods, loading all BP mods in random order."))
+        Log(string.format("Mods/BPModLoaderMod/load_order.txt not present or no matching mods, loading all BP mods in random order.\n"))
     end
 end
 
@@ -124,9 +124,9 @@ local function LoadModConfigs()
         LogicModsDir = Dirs.Game.Content.Paks.LogicMods
         if not LogicModsDir then error("[BPModLoader] Unable to find or create Content/Paks/LogicMods directory. Try creating manually.") end
     end
-    for ModDirectoryName,ModDirectory in pairs(LogicModsDir) do
+    for ModDirectoryName, ModDirectory in pairs(LogicModsDir) do
         Log(string.format("Mod: %s\n", ModDirectoryName))
-        for _,ModFile in pairs(ModDirectory.__files) do
+        for _, ModFile in pairs(ModDirectory.__files) do
             Log(string.format("    ModFile: %s\n", ModFile.__name))
             if ModFile.__name == "config.lua" then
                 dofile(ModFile.__absolute_path)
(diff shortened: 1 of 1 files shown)


UE4SS:
> how it can be that `ExistingActor:GetWorld()` return a `nil`.

This is how:
```c++
// UE4SS/src/LuaType/LuaAActor.cpp -> setup_member_functions -> GetWorld
if (Unreal::UWorld* world = lua_object.get_remote_cpp_object()->GetWorld(); world)
{
    auto_construct_object(lua, world);
}
else
{
    lua.set_nil();
}
```
I consider it a bug to return `nil` here, the entire if-statement should be deleted.
The same is true for `GetLevel`.

igromanru:
> > how it can be that `ExistingActor:GetWorld()` return a `nil`.
> 
> This is how:
> 
> ```c++
> // UE4SS/src/LuaType/LuaAActor.cpp -> setup_member_functions -> GetWorld
> if (Unreal::UWorld* world = lua_object.get_remote_cpp_object()->GetWorld(); world)
> {
>     auto_construct_object(lua, world);
> }
> else
> {
>     lua.set_nil();
> }
> ```
> 
> I consider `lua.set_nil()` to be a bug here, the entire if-statement should be deleted, leaving only `auto_construct_object`. The same is true for `GetLevel`.

Agree. Should it be another PR then? What is with current changes?

UE4SS:
> > > how it can be that `ExistingActor:GetWorld()` return a `nil`.
> > 
> > 
> > This is how:
> > ```c++
> > // UE4SS/src/LuaType/LuaAActor.cpp -> setup_member_functions -> GetWorld
> > if (Unreal::UWorld* world = lua_object.get_remote_cpp_object()->GetWorld(); world)
> > {
> >     auto_construct_object(lua, world);
> > }
> > else
> > {
> >     lua.set_nil();
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I consider `lua.set_nil()` to be a bug here, the entire if-statement should be deleted, leaving only `auto_construct_object`. The same is true for `GetLevel`.
> 
> Agree. Should it be another PR then? What is with current changes?

Frankly I think this PR is in an unacceptable state.
You've included a bunch of formatting changes which makes it much harder to see what the actual change is.
If the formatting changes are intentional, feel free to make a new PR specifically for that.
I'm still fine with making it a warning instead of an error, because it will still be very helpful when diagnosing problems, but you need to clean up the PR.

However, the proper fix to this problem is to fix GetWorld/GetLevel in C++, and I don't mind it being in the same PR but it needs its own entry in the changelog.

igromanru:
As stated in the commit message, I formatted the file, to make it overall better readable.  
I don't see what's wrong with it or why I should make another RP just to fix formatting.  
It's basically just another RP to improve BPModLoaderMod as follow up to #752.  
I would rather make a new RP to fix the issue that `GetWorld()` can return `nil`, because it's overall a different issue that has an indirect impact on the BPModLoaderMod.

UE4SS:
> As stated in the commit message, I formatted the file, to make it overall better readable. I don't see what's wrong with it or why I should make another RP just to fix formatting. It's basically just another RP to improve BPModLoaderMod as follow up to #752. I would rather make a new RP to fix the issue that `GetWorld()` can return `nil`, because it's overall a different issue that has an indirect impact on the BPModLoaderMod.

I don't want to fish through unrelated widespread changes to find out what the actual code change is, I'm more likely to delay looking at the PR or even ignore it completely instead.
The only commit in this PR has +17/-14 changes, and only two of those actually matter.
I will make an exception this time, but for the future, I would find it more acceptable if it was at least a second commit.

UE4SS (approved):

Lyrth:
I'm not in favor of making exceptions, as this will exist permanently in history (unless rewritten or reverted of course). Formatting should not consist like >70% of a commit if the commit is meant for something else -- I agree with it being a second (`chore:`) commit, but at that point might as well separately perform the formatting on other Lua files too for consistency.

igromanru:
Jesus Christ. You guys can't make up your mind.  
Discord_H4pPWQb00q  
Discord_5ZS0MfnU1c  
What do you want me to do? Wait until the PR is approved first to squash my commits?  
So it has to be approved a second time? A new branch and RP for each refactoring and formatting?  
I guess no PRs, no problems...

UE4SS:
> What do you want me to do? Wait until the PR is approved first to squash my commits? So it has to be approved a second time? A new branch and RP for each refactoring and formatting? I guess no PRs, no problems...

You should never indiscriminately squash everything down into a single commit.
If you have WIP commits, those make sense to squash, but otherwise it makes more sense that the person merging does the squashing.
What we want are clean PRs, if you can't figure out what a clean PR looks like then feel free to leave all your WIP commits and we can squash them for you later.

As far as the screenshot is concerned, Lyrthras doesn't make the rules here, nor are they responsible for merging anything, so while it's nice for us if people make clean PRs, feel free to ignore what they said and we clean it up for you.

> I'm not in favor of making exceptions, as this will exist permanently in history (unless rewritten or reverted of course). Formatting should not consist like >70% of a commit if the commit is meant for something else -- I agree with it being a second (`chore:`) commit, but at that point might as well separately perform the formatting on other Lua files too for consistency.

Noted. However, this is a one-time exception, it won't become a habit.

PR #814, opened 2025-02-27
fix: opened object in live view no longer gets clipped early when scr…

Corporalwill123:
**Type of change**

**Checklist**

### DIFF
--- a/UE4SS/src/GUI/LiveView.cpp
+++ b/UE4SS/src/GUI/LiveView.cpp
@@ -3596,10 +3596,29 @@ namespace RC::GUI
             // 2) Use clipper with the filtered array size
             ImGuiListClipper clipper{};
             clipper.Begin(objects_to_draw.size(), ImGui::GetTextLineHeightWithSpacing());
+
+            // Forces the current opened node to always be rendered by the clipper
+            for (int i = 0; i < objects_to_draw.size(); i++)
+            {
+                if (objects_to_draw[i] == m_currently_opened_tree_node)
+                {
+                    clipper.ForceDisplayRangeByIndices(i, i + 1);
+                    break;
+                }
+            }
+            int last_display_end = 0;
+            float last_position = ImGui::GetCursorPosY();
             while (clipper.Step())
             {
-
+                if (last_position > ImGui::GetCursorPosY())
+                {
+                    // Makes sure the list is contiguous
+                    ImGui::SetCursorPosY(last_position);
+                    clipper.DisplayStart = last_display_end;
+                }
                 do_iteration(clipper.DisplayStart, clipper.DisplayEnd, &objects_to_draw);
+                last_position = ImGui::GetCursorPosY();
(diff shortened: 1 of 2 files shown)


UE4SS:
This patch works for me.

To test:
1. Open Live Viewer and select an object.
2. Drill down into the Super/Class hierarchy until it covers a lot of the live view area.
3. Scroll up/down.

Before the patch, it would suddenly clip the entire thing when scrolling up/down a small amount.
After the patch, it only clips after it's no longer in the visible area, making the experience a lot smoother.

Corporalwill123:
Here is some clarification on this PR:

**Problem**
Here is a class with lots of properties as an example, so it's obvious when the problem occurs
image
now if we scroll down a tiny bit to view more of the properties, this happens:
image
The opened object gets immediately clipped off once the root of the selected node passes the top edge

**Cause**
By default, Dear ImGUI will automatically create scrollbars when the contents of a window don't fit, and so you can just make a draw call for each item and it will handle the rest.
Usually this is fine, but with large lists there will be a large amount of draw calls which will be wasted, since most of them will not display to the window.

This is when the ImGUIListClipper comes in, which will take a the number of list items and the size of each item to figure out the range of items to display, so only the items within the region will have draw calls running.

For regular lists this is fine, but since LiveView uses an expandable tree, one item will have a different size compared to the others, so this happens
image
In the image, the clipped items are red, the unclipped/succesfully drawn items are green, the wasted draw calls are orange, the expanded node is blue, and the white rectangle is the viewport
On the left is what ImGUIListClipper predicts, the right is what gets displayed.

**Solution**
To fix this, I use ImGUIListClipper's *ForceDisplayRangeByIndices* function to make sure the expanded item's draw calls always run.
This can cause the while loop to run more than once.
Now since the clipper always outputs ranges in order from lowest to highest index, and combines ranges when they overlap, when the expanded node is within the clipped region, it runs as normal and loops only once, if it is above, it gets ran before the clipped region, and if it is below, it gets ran after.

Then what we do, is if the starting cursor position is above the last iteration's position, we adjust the cursor to bottom of the last position, and also make sure to extend the range to include all the items in between.
We extend the range, because if you notice in the diagram, in the Expected viewport after scrolling, there will be items between the expanded node and the first non-clipped item, so we have to add those back in.

The reason this works is because only a single item is allowed to be expanded at a time, if there were multiple expanded at the same time, then there would need to be a more complicated fix.

So in the end, if the expanded node is
- above the viewport: the draw call is wasted, and the rest gets displayed as normal.
- intersecting the top of the viewport: it gets drawn properly, and all subsequent calls get adjusted to keep the list contiguous, wasting draw calls at the end as expected.
- in the center: nothing special happens, wasted calls at the end of the viewport like in original behavior
- below the bottom of the viewport: the normal drawing happens, and then the expanded node wastes its draw calls

UE4SS (approved):

PR #817, opened 2025-02-28
Fix proxy_generator proxy being unbuildable due to uninitialized variable

Lyrth:
## Description

Fixes "fatal error LNK1121: duplicate ordinal number '1'" when building the proxy, due to undefined behavior of `ordinal` variable if not set nor initialized.

## Type of change

## How Has This Been Tested?

A user that had above fatal error on build has reported that the proxy has succeeded building after adding the change.

### DIFF
--- a/UE4SS/proxy_generator/main.cpp
+++ b/UE4SS/proxy_generator/main.cpp
@@ -112,7 +112,7 @@ std::vector<ExportFunction> ReadExportsFile(const fs::path& exp_path, fs::path& 
     {
         std::istringstream s(line);
 
-        uint16_t ordinal;
+        uint16_t ordinal{};
         string export_name;
         s >> ordinal >> export_name;
 


igromanru:
It has no bearing on the PR, but I personally dislike the "lazy" approach that Martin suggested to initialize everything with `{}`.  
Pretty sure that the compiler will optimize it himself, but `{}` usually results into a `memset`, while assigning a simple integer value is just `mov ax, 0` or in this case it would be `xor ax, ax`.  
I just don't see the gain to use `{}` instead of `= 0` when you know that it's an `uint16_t` and find it ugly.  
But this is just my opinion.

UE4SS (approved):

narknon:
Compilers are smart enough to not use memset for bracket initialization of basic types. Compiler may use memset on a large array or struct, but manually initializing each member is way more error prone, particularly if the struct is ever changed in the future, so bracket initializer still has a benefit in most cases.

PR #823, opened 2025-03-06
Added ProcessLocalScriptFunction Override functionality

M3C3I:
This change requires changes in the Unreal deps as well for this override to function.

**Description**

Added ability to use a lua script to override ProcessLocalScriptFunction

**Type of change**

- [X] New feature (non-breaking change which adds functionality)
- [X] Is/requires documentation update
    - Potentially could be added to [this page](https://docs.ue4ss.com/guides/fixing-compatibility-problems.html)

**How Has This Been Tested?**

- Created the file ProcessLocalScriptFunction.lua in ue4ss\UE4SS_Signatures which successfully overrides the built-in method.
- Removed the ProcessLocalScriptFunction.lua from that same directory and it successfully uses the built-in method.

**Checklist**

- [x] Pull Request [HERE](https://github.com/Re-UE4SS/UEPseudo/pull/123) will need to be merged before this can be merged.

### DIFF
--- a/UE4SS/src/Signatures.cpp
+++ b/UE4SS/src/Signatures.cpp
@@ -323,7 +323,7 @@ namespace RC
                 scan_from_lua_script(
                         lua_process_local_script_function_scan_script,
                         signature_containers,
-                        [&scan_result](void* address) {
+                        [](void* address) {
                             Output::send(STR("ProcessLocalScriptFunction address: {} <- Lua Script\n"), address);
                             Unreal::UObject::ProcessLocalScriptFunctionInternal.assign_address(address);
                             return DidLuaScanSucceed::Yes;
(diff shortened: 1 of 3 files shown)


UE4SS/src/Signatures.cpp
```diff
@@ -315,5 +315,48 @@ namespace RC
...
+        auto lua_process_local_script_function_scan_script = working_directory / "UE4SS_Signatures/ProcessLocalScriptFunction.lua";
+        if (std::filesystem::exists(lua_process_local_script_function_scan_script))
+        {
+            config.ScanOverrides.process_local_script_function = [lua_process_local_script_function_scan_script](std::vector<SignatureContainer>& signature_containers,
+                                                                                              Unreal::Signatures::ScanResult& scan_result) mutable {
+                scan_from_lua_script(
+                        lua_process_local_script_function_scan_script,
+                        signature_containers,
+                        [](void* address) {
+                            Output::send(STR("ProcessLocalScriptFunction address: {} <- Lua Script\n"), address);
```
UE4SS, line 327: Your UEPseudo PR does this as well so won't this be output twice ?
Same goes for ProcessInternal.

M3C3I: I put it in both spots originally to make sure the Lua scan addresses weren’t getting overwritten by the UEPseudo if-block. Now that I’ve tested it and the overrides work fine, I’d say we should drop the output from the Lua scans. UEPseudo still does more processing in UnrealInitializer.cpp before finalizing the address, so it feels like the better place to keep it.

I'll make the commit to remove the Output lines from here and keep them only in the UEPseudo. If you would prefer it a different way, let me know.

UE4SS: Well the UEPseudo repo is a self-contained code-base so it's a bit strange to have mentions of Lua there.

The UEPseudo code-base isn't aware why or from where the override has happened, just that it happened.
Someone could come along later and override it in C++ and it'll incorrectly display "Lua".
Besides, we have the rest of the override output in UE4SS not in UEPseudo so may as well continue that.

We do unfortunately have some other mentions of Lua in UEPseudo that snuck in at some point but that doesn't make introducing further ones any better.

I would recommend undoing the latest commit here, and moving the output in UEPseudo to within the if-statement checking that it's not overridden and obviously removing the Lua part of the message.

M3C3I: Okay. That makes sense to keep things consistent. I will make the changes.

M3C3I: @UE4SS I have updated the [UEPseudo Pull Request](https://github.com/Re-UE4SS/UEPseudo/pull/123) and the Signatures.cpp here.
Let me know if there is anything else that is required of me before this can be merged.

M3C3I: @UE4SS This can now be merged once the UEPseudo is merged. I had previously forgot to put the static cast in the ouput message in UEPseudo which was causing a build error. Fixed now and everything on the latest commits have been tested.

Let me know if there is anything else that is required of me before this can be merged.

UE4SS (approved):

PR #831, opened 2025-03-08
Update issue templates

Buckminsterfullerene02:
**Description**

- Added new docs report & feature request issue templates
- Bug report and PR templates updated to put instructions into markdown comments so they don't need to be manually deleted (or left in which reads badly)
- Templates are auto-labelled on what they are

**Type of change**

- [x] Other... Please describe: Repo management

**How Has This Been Tested?**

It hasn't aside from checking the markdown is valid with preview mode and checking with the docs that the metadata is valid (e.g. labels with spaces in working as they are)

**Checklist**

N/A

**Additional context**


UE4SS (approved):
This seems like a good change.

PR #835, opened 2025-03-10
UEHelpers: Fix GetActorFromHitResult and refactor it a little

igromanru:
**Description**
I found an embarrassing error that I didn't test properly.  
`FHitResult` is a struct and has no `IsValid()` function.

**Changes**
- Removed HitResult:IsValid() check. It's a struct and has no such function.
- Changed the description of GetActorFromHitResult to be slightly more accurate, as it doesn't just deal with the difference between UE4 and UE5
- Removed `---@type AActor` annotation from `return CreateInvalidObject()`. It returns an UObject, which is a valid return type for GetActorFromHitResult.

**Type of change**

**How has this been tested?**
Tested in my Line Trace mod.

### DIFF
--- a/assets/Mods/shared/UEHelpers/UEHelpers.lua
+++ b/assets/Mods/shared/UEHelpers/UEHelpers.lua
@@ -195,12 +195,12 @@ function UEHelpers.GetAllPlayers()
 end
 
 ---Returns hit actor from FHitResult.<br>
----The function handles the struct differance between UE4 and UE5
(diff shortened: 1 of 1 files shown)


igromanru:
It would be nice if the PR could be merged soon, I would like to release current state of UE4SS to the NM pages that I maintain.

UE4SS (approved):

igromanru:
Thanks

PR #864, opened 2025-04-25
fix: pass scale to UEPseudo UWorld::SpawnActor

massclown:
**Description**
Currently in UE4SS it is not easy to spawn scaled actors from Lua, and for some actors setting scale after spawning may not work (animated characters, etc.).

In [Re-UE4SS/UEPseudo/blob/main/src/World.cpp](https://github.com/Re-UE4SS/UEPseudo/blob/main/src/World.cpp), [the first implementation](https://github.com/Re-UE4SS/UEPseudo/blob/3185035434260bfd82224375f31043d18e362cac/src/World.cpp#L15) of `UWorld::SpawnActor` hardcodes the scale to 1.0:
```
    AActor* UWorld::SpawnActor(UClass* InClass, FVector const* Location, FRotator const* Rotation)
    {
        Unreal::FTransform Transform{
                Rotation ? FQuat(*Rotation) : FQuat(),
                Location ? *Location : FVector(),
                {1.0f, 1.0f, 1.0f},
...
```
and [the second](https://github.com/Re-UE4SS/UEPseudo/blob/3185035434260bfd82224375f31043d18e362cac/src/World.cpp#L32) `AActor* UWorld::SpawnActor(UClass* Class, FTransform const* Transform)` needs an `FTransform` which is not passed through Lua currently.

To follow the style of the existing API, we add another Lua overload, that also accepts one more, the fourth table parameter, a FVector-like table of scale:
```
SpawnActor(UClass ActorClass, FVector Location, FRotator Rotation, FVector Scale)
```
and constructs an `FTransform` to call the relevant second implementation from UEPseudo.

The original `SpawnActor(UClass ActorClass, FVector Location, FRotator Rotation)` should still work as before.

**Type of change**

**How has this been tested?**
Built locally and tested manually.

### DIFF
--- a/UE4SS/src/LuaType/LuaUWorld.cpp
+++ b/UE4SS/src/LuaType/LuaUWorld.cpp
@@ -65,7 +65,8 @@ namespace RC::LuaType
             std::string error_overload_not_found{R"(
 No overload found for function 'SpawnActor'.
 Overloads:
-#1: SpawnActor(UClass Class, table Location, table Rotation))"};
+#1: SpawnActor(UClass Class, table Location, table Rotation)
+#2: SpawnActor(UClass Class, table Location, table Rotation, table Scale))"};
 
             if (!lua.is_userdata())
             {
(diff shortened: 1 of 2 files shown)


UE4SS:
This breaks from UEs API.
There is no SpawnActor overload in UE that takes scale directly as a parameter.

An alternative is to expose the overload that takes FTransform to Lua.
What do you think @narknon ?

narknon:
> This breaks from UEs API. There is no SpawnActor overload in UE that takes scale directly as a parameter.
> 
> An alternative is to expose the overload that takes FTransform to Lua. What do you think @narknon ?

I  haven't reviewed the code but it seems relatively benign if we could have an extra overload that just has scale without breaking the others and with still allowing us to have an FTransform version.  I guess it's possible it could cause issues in the future if UE does another overload that somehow won't work with our overloads, but that seems unlikely.

massclown:
I can take a stab at implementing the FTransform API with a table of tables like: 
```
SpawnActor(UClass Class, table Transform {Location={X,Y,Z}, Rotation={Pitch,Yaw,Roll}, Scale={X,Y,Z}}))"};
```
Would that be better?

UE4SS:
> I can take a stab at implementing the FTransform API with a table of tables like:
> 
> ```
> SpawnActor(UClass Class, table Transform {Location={X,Y,Z}, Rotation={Pitch,Yaw,Roll}, Scale={X,Y,Z}}))"};
> ```
> 
> Would that be better?

I would prefer it, but I'm not gonna block the original idea that just adds the Scale3D param, up to you to decide now.

If you do go ahead with the `FTransform` overload, here are a few problems:
The rotation for `FTransform` is an `FQuat` not an `FRotator`.
Your order is also incorrect, which I think you should correct even if it only affects the docs.
The correct order is `Rotation`, `Translation`, `Scale3D`.

PR #866, opened 2025-04-29
UE4SS_ObjectDump: Add UFunction::Func, and re-order functions to not be in a random place in the file

### LINKED ISSUE #865: How i can get the function offset relative to the base address  from dumpobjects
**Is your feature request related to a problem? Please describe.**

For an Unreal Engine game, after dumping the objects, I noticed that the data lacks base address-relative offsets. This makes it inconvenient to locate corresponding functions in IDA for analysis. To address this, could the dumped data include the offsets relative to the current EXE's base address? This way, by directly jumping to these offsets in IDA, function hooking would become significantly more efficient

**Describe the solution you'd like**

Key points to emphasize in the translation:

Missing base-relative offsets in dumped objects.
Difficulty in IDA navigation due to the absence of offsets.
Request for offset integration in dump outputs to streamline function hooking.
This translation preserves the technical context while clarifying the need for base-address offsets in UE reverse engineering workflows.

UE4SS:
**Description**

This PR adds `f` to UE4SS_ObjectDump.txt which represents the function pointer to either the native function wrapper, or to ProcessInternal.
This PR also makes sure that functions appear immediately under class properties instead whatever random order they happened to have in GUObjectArray.

Fixes #865 

**Type of change**

**How has this been tested?**

Generate a dump from the same game, at the exact same state, and compare the two files to make sure the number of lines match.

**Checklist**

### DIFF
--- a/assets/Changelog.md
+++ b/assets/Changelog.md
@@ -47,6 +47,8 @@ Added override Lua files for CallFunctionByNameWithArguments [UE4SS #848](https:
 
 Add error messages in places where only error codes were previously logged (e.g. load a C++ mod) [UE4SS #902](https://github.com/UE4SS-RE/RE-UE4SS/pull/902) 
 
+Added `[f: <address_or_module_offset>` section to UE4SS_ObjectDump.txt [UE4SS #866](https://github.com/UE4SS-RE/RE-UE4SS/pull/866) 
+
 ### Live View 
 Added search filter: `IncludeClassNames`. ([UE4SS #472](https://github.com/UE4SS-RE/RE-UE4SS/pull/472)) - Buckminsterfullerene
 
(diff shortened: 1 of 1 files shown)


UE4SS:
Don't merge this yet!
There are problems in UE versions before 4.25, working on fixing those.

UE4SS:
This PR is now ready to be tested for both <4.25 and 4.25+.

UE4SS:
Needs further work: https://github.com/UE4SS-RE/RE-UE4SS/issues/865#issuecomment-2889017574

UE4SS:
The creator of #865 is unresponsive so I'm moving ahead with this PR and we can deal with any potential problems later.

PR #882, opened 2025-05-19
Feature: Live View search by memory address

vitrvvivs:
**Description**
Add logic to the LiveView search to treat hex numbers as memory addresses, and return the object that probably contains it (reusing the size calculation from elsewhere; maybe inaccurate).
Disabled by default. Few people might use it, and it does add some time to the search (but not perceptibly on my machine).

**Type of change**

**How has this been tested?**
Only tested on UE 5.3 (Oblivion)
- find the address of an object, then search for it

**Checklist**

### DIFF
--- a/UE4SS/src/GUI/LiveView.cpp
+++ b/UE4SS/src/GUI/LiveView.cpp
@@ -177,6 +177,23 @@ namespace RC::GUI
         }
     }
 
+    static void attempt_to_add_search_by_address_result(uintptr_t address_to_search_by, UObject* object)
+    {
+        auto object_addr = std::bit_cast<uintptr_t>(object);
+        if (address_to_search_by < object_addr)
+        {
+            return;
+        }
+
+        auto uclass = object->IsA<UStruct>() ? static_cast<UClass*>(object) : object->GetClassPrivate();
+        uintptr_t end_addr = object_addr + uclass->GetPropertiesSize();
+        //Output::send<LogLevel::Default>(STR("object_addr {:016X}-{:016X}"), object_addr, end_addr);
+        if (address_to_search_by < end_addr) {
+            LiveView::s_name_search_results.emplace_back(object);
+            LiveView::s_name_search_results_set.emplace(object);
+        }
+    }
+
     static auto remove_search_result(UObject* object) -> void
     {
         LiveView::s_name_search_results.erase(std::remove_if(LiveView::s_name_search_results.begin(),
@@ -1817,8 +1834,25 @@ namespace RC::GUI
         Output::send(STR("Searching by name...\n"));
         s_name_search_results.clear();
         s_name_search_results_set.clear();
+
+        uintptr_t address_to_search_by = 0;
+        if (UE4SSProgram::settings_manager.General.SearchByAddress &&
+           (LiveView::s_name_to_search_by.size() >= 6 && LiveView::s_name_to_search_by.size() <= 16))
+        {
+            try
+            {
+                address_to_search_by = std::stoull(LiveView::s_name_to_search_by, nullptr, 16);
+                Output::send<LogLevel::Default>(STR("Searching for object with address: {:016X}"), address_to_search_by);
+            }
+            catch (...) {}
+        }
+
         UObjectGlobals::ForEachUObject([&](UObject* object, ...) {
             attempt_to_add_search_result(object);
+            if (address_to_search_by)
+            {
+                attempt_to_add_search_by_address_result(address_to_search_by, object);
+            }
             return LoopAction::Continue;
         });
     }
--- a/assets/UE4SS-settings.ini
+++ b/assets/UE4SS-settings.ini
@@ -23,6 +23,10 @@ SecondsToScanBeforeGivingUp = 30
 ; Default: true
 bUseUObjectArrayCache = true
 
+; When the search query in the Live View tab is a hex number, try to find the UObject that contains that memory address.
+; Default: false
+bEnableSeachByMemoryAddress = false
+
 [EngineVersionOverride]
 MajorVersion = 
 MinorVersion = 
--- a/UE4SS/include/SettingsManager.hpp
+++ b/UE4SS/include/SettingsManager.hpp
@@ -26,6 +26,7 @@ namespace RC
             int64_t SecondsToScanBeforeGivingUp{30};
             bool UseUObjectArrayCache{true};
             StringType InputSource{STR("Default")};
+            bool SearchByAddress{false};
         } General;
 
         struct SectionEngineVersionOverride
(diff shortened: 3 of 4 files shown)


UE4SS/src/GUI/LiveView.cpp
```diff
@@ -1817,8 +1834,25 @@ namespace RC::GUI
         Output::send(STR("Searching by name...\n"));
         s_name_search_results.clear();
         s_name_search_results_set.clear();
+
+        uintptr_t address_to_search_by = 0;
+        if (UE4SSProgram::settings_manager.General.SearchByAddress &&
+           (LiveView::s_name_to_search_by.size() >= 6 && LiveView::s_name_to_search_by.size() <= 16))
```
UE4SS, line 1840: Is it necessary to check the size ? Doesn't `std::stoull` throw an exception if it's not valid anyway ?

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -1817,8 +1834,25 @@ namespace RC::GUI
...
+        {
+            try
+            {
+                address_to_search_by = std::stoull(LiveView::s_name_to_search_by, nullptr, 16);
+                Output::send<LogLevel::Default>(STR("Searching for object with address: {:016X}"), address_to_search_by);
+            }
+            catch (...) {}
```
UE4SS, line 1847: Let's not fail silently.
Could you add a popup window so that the user is made aware immediately of the error.
We do that if a user tries to open a nullptr object, and you can search for `PropertyOpenNullptrWarning` for reference.

assets/UE4SS-settings.ini
```diff
@@ -23,6 +23,10 @@ SecondsToScanBeforeGivingUp = 30
 ; Default: true
 bUseUObjectArrayCache = true
 
+; When the search query in the Live View tab is a hex number, try to find the UObject that contains that memory address.
+; Default: false
+bEnableSeachByMemoryAddress = false
```
UE4SS, line 28: This doesn't seem like an appropriate place for this option.
I think it makes more sense to add it to the search options that appear when you right-click the search bar.
And then you'll also need to update the docs in [docs/feature-overview/live-view.md](https://github.com/UE4SS-RE/RE-UE4SS/blob/main/docs/feature-overview/live-view.md) to explain this feature.
image

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -1817,8 +1834,25 @@ namespace RC::GUI
...
+        }
+
         UObjectGlobals::ForEachUObject([&](UObject* object, ...) {
             attempt_to_add_search_result(object);
```
UE4SS, line 1851: Do we really want to do a name search even if we're also doing an address search ?
I fail to see you can search for both at the same time.

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -177,6 +177,23 @@ namespace RC::GUI
...
+    {
+        auto object_addr = std::bit_cast<uintptr_t>(object);
+        if (address_to_search_by < object_addr)
+        {
+            return;
+        }
+
+        auto uclass = object->IsA<UStruct>() ? static_cast<UClass*>(object) : object->GetClassPrivate();
+        uintptr_t end_addr = object_addr + uclass->GetPropertiesSize();
+        //Output::send<LogLevel::Default>(STR("object_addr {:016X}-{:016X}"), object_addr, end_addr);
```
UE4SS, line 190:
```suggestion
```
Remove this debug message please.

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -177,6 +177,23 @@ namespace RC::GUI
...
+        if (address_to_search_by < end_addr) {
```
UE4SS, line 191:
```suggestion
        if (address_to_search_by <= end_addr) {
```
Shouldn't this be inclusive ? If the address matches exactly then the result is probably still something that the user is interested in.

UE4SS (changes requested):

dicene:
I'll be excited to have this functionality once it's in.

UE4SS/src/GUI/LiveView.cpp
vitrvvivs: Yeah, this is a remnant from trying to speed it up, that I didn't rethink. (`stoull` is relatively expensive, and I had originally put it inside the loop, in the `attempt_to_add_search_result` function.) Unnecessary now.

UE4SS/src/GUI/LiveView.cpp
vitrvvivs: There could be some extra fields not reported in the size, or padding. I don't know much about unreal internals. In my testing there was some padding, but I don't know how consistent that is.
`end_addr` is probably misnamed. If it were in a C array, and `GetPropertiesSize()` is accurate, then address+size = the address of the next object.

assets/UE4SS-settings.ini
vitrvvivs: I figured it would be UI clutter for something most people wouldn't use.

UE4SS/src/GUI/LiveView.cpp
vitrvvivs: The question (RE: a few comments) is whether to handle it transparently or explicitly. In most cases, a 64-bit address won't match any object names (unless the devs did something funny), and a search-by-name that successfully parses as a number (itself relatively rare) won't match any addresses. Requiring the user to specify which one they're searching for, when the program can figure it out, seems unnecessary. And the best way to figure out whether it's a name that happens to be a number, or an address, is to try both.

assets/UE4SS-settings.ini
UE4SS: Most people don't use any of the options, so I don't think one more option is a big deal.

UE4SS/src/GUI/LiveView.cpp
```diff
@@ -1817,8 +1840,24 @@ namespace RC::GUI
...
+
+        uintptr_t address_to_search_by = 0;
+        if (LiveView::s_search_by_address)
+        {
+            try
+            {
+                address_to_search_by = std::stoull(LiveView::s_name_to_search_by, nullptr, 16);
+            }
+            catch (std::invalid_argument) {} // query isn't a valid hex number; can't be an address
+            catch (std::out_of_range) {} // is a hex number, but too big, or negative; can't be address
```
UE4SS, line 1852: We're silently failing here.
I'd prefer if this was a blocking popup box containing the error message, but will settle for `Output::send<LogLevel::Error>`.

UE4SS (changes requested):

UE4SS:
I apologize for the extreme delay, I didn't get any notification for your last two commits.

UE4SS/src/GUI/LiveView.cpp
UE4SS: > I'm interested in getting this feature merged, so I made this change locally. Is there a preferred way for me to contribute those changes to this PR? e.g. open a new PR, PR into this PR's fork, attach a patch file for a follow-up PR

First step is to push it to your own fork so we can look at it.
We might be able to cherry pick your commit(s).

ntsik: Diff is here: https://github.com/vitrvvivs/RE-UE4SS/compare/search-by-memory-address...ntsik:RE-UE4SS:search-by-memory-address

The base is out-of-date, so the last change doesn't have much context around it in the diff. Off of `main`, I put it in `LiveView::render` underneath the other modals.

UE4SS: Thanks. I've cherry picked your commit and rebased the whole thing onto latest main.

UE4SS (approved):

PR #883, opened 2025-05-20
feat(lua): adds support for `FSetProperty` in Lua scripting

narknon:
**Description**
Adds Lua bindings for `FSetProperty`, allowing access to Unreal Engine sets from Lua scripts.

This includes:
- Adding `FScriptSetLayout` to the list of valid UDT names.
- Updating member variable layout templates for various engine versions to include `SetLayout` offset.
- Adding Lua type definitions and functions for `TSet`.
- Adding a pusher for `SetProperty` to the Lua state.

**Type of change**

**How has this been tested?**

Tested via the following lua script. This test will only work on UE versions after the ConcurrencySet property was added (I believe sometime after 4.26), and would need to be modified for other versions.
[TSetTest.zip](https://github.com/user-attachments/files/20358105/TSetTest.zip)

**Checklist**

    - Requires Re-UE4SS/UEPseudo#129


UE4SS (commented):
Can you clarify where `SetLayout` in `FSetProperty` came from ?
It wasn't there before obviously, and this PR doesn't seem to change UVTD to add it.
Was this method previously manually deleted from files before being pushed to the repo or something ?

narknon:
> Can you clarify where `SetLayout` in `FSetProperty` came from ? It wasn't there before obviously, and this PR doesn't seem to change UVTD to add it. Was this method previously manually deleted from files before being pushed to the repo or something ?

assets/Default_UVTD_Configs/Config/valid_udt_names.json

UE4SS:
> > Can you clarify where `SetLayout` in `FSetProperty` came from ? It wasn't there before obviously, and this PR doesn't seem to change UVTD to add it. Was this method previously manually deleted from files before being pushed to the repo or something ?
> 
> assets/Default_UVTD_Configs/Config/valid_udt_names.json

This PR only adds `FScriptSetLayout` to that file.
It already had `FSetProperty`, but that doesn't explain why `FSetProperty::SetLayout` is magically appearing now.

narknon:
> This PR only adds `FScriptSetLayout` to that file. It already had `FSetProperty`, but that doesn't explain why `FSetProperty::SetLayout` is magically appearing now.

Because that is the return type so it wasn't dumped as a valid udt type prior.

narknon:
> > This PR only adds `FScriptSetLayout` to that file. It already had `FSetProperty`, but that doesn't explain why `FSetProperty::SetLayout` is magically appearing now.
> 
> Because that is the return type so it wasn't dumped as a valid udt type prior.

UVTD is quite confusing. I should get back to my refactor.

Lyrth:
Tested, changes look good to me.

Should remember to also update Types.lua

narknon:
Could you make an issue for the TArray thing you mentioned?

Lyrth:
> Could you make an issue for the TArray thing you mentioned?

It appears that the issue involves a wider scope than just TArray, should probably be decided at a later time instead.

narknon:
> It appears that the issue involves a wider scope than just TArray, should probably be decided at a later time instead.

An issue for the aspects we discussed about table with member vs not, etc. would still be helpful for tracking.

Lyrth:
> An issue for the aspects we discussed about table with member vs not, etc. would still be helpful for tracking.

Done

PR #886, opened 2025-05-22
Sanitize some of the BPmodloader logic

Kein:
We shouldn't do anything if there is no PAK mods.

### DIFF
--- a/assets/Mods/BPModLoaderMod/Scripts/main.lua
+++ b/assets/Mods/BPModLoaderMod/Scripts/main.lua
@@ -1,24 +1,43 @@
+---########################
+--- DEFINES
+---########################
+package.path = '.\\Mods\\ModLoaderMod\\?.lua;' .. package.path
+package.path = '.\\Mods\\ModLoaderMod\\BPMods\\?.lua;' .. package.path
 local UEHelpers = require("UEHelpers")
 
 local VerboseLogging = false
+local SpawnModsOnLuaInit = true -- spawn on lua unit exec? common cause of random crashes (game-specific race?)
+local AssetRegistryHelpers = CreateInvalidObject() ---@cast AssetRegistryHelpers UAssetRegistryHelpers
+local AssetRegistry = CreateInvalidObject() ---@cast AssetRegistry IAssetRegistry
+
+local Mods = {}
+local OrderedMods = {}
+local ModOrderList = {} -- contains entries from Mods/BPModLoaderMod/load_order.txt; used for the load order of BP mods.
+local DefaultModConfig = {
+    AssetName = "ModActor_C",
+    AssetNameAsFName = UEHelpers.FindOrAddFName("ModActor_C")
+}
+
+
+---########################
+--- HELPERS
+---########################
 
 local function Log(Message, OnlyLogIfVerbose)
     if not VerboseLogging and OnlyLogIfVerbose then return end
     print("[BPModLoaderMod] " .. Message)
 end
 
-package.path = '.\\Mods\\ModLoaderMod\\?.lua;' .. package.path
-package.path = '.\\Mods\\ModLoaderMod\\BPMods\\?.lua;' .. package.path
-
-local Mods = {}
-local OrderedMods = {}
-
--- Contains mod names from Mods/BPModLoaderMod/load_order.txt and is used to determine the load order of BP mods.
-local ModOrderList = {}
-
-local DefaultModConfig = {}
-DefaultModConfig.AssetName = "ModActor_C"
-DefaultModConfig.AssetNameAsFName = UEHelpers.FindOrAddFName("ModActor_C")
+--- get table count, never rely on #t
+--- @param T table
+--- @return integer
+local function GetModCount(T)
+    local c = 0
+    for _, ModInfo in ipairs(OrderedMods) do
+        if type(ModInfo) == "table" then c = c +1 end
+    end
+    return c
+end
 
 -- Checks if the beginning of a string contains a certain pattern.
 local function StartsWith(String, StringToCompare)
@@ -47,6 +66,17 @@ local function LinesFrom(file, ignoreLinesStartingWith)
         end
     end
     return lines
+end
+
+local function LogOrderedMods()
+    for _, v in ipairs(OrderedMods) do
+        Log(string.format("%s == %s\n", v.Name, v))
+        if type(v) == "table" then
+            for k2, v2 in pairs(v) do
+                Log(string.format("    %s == %s\n", k2, v2))
+            end
+        end
+    end
 end
 
 -- Loads mod order data from load_order.txt and pushes it into ModOrderList.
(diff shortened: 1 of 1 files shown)


narknon:
I think this looks good but would like @UE4SS or someone that works with Lua more to look at it.

assets/Mods/BPModLoaderMod/Scripts/main.lua
```diff
@@ -110,26 +140,49 @@ local function SetupModOrder()
...
+        return
     end
+
     for ModDirectoryName, ModDirectory in pairs(LogicModsDir) do
         Log(string.format("Mod: %s\n", ModDirectoryName))
         for _, ModFile in pairs(ModDirectory.__files) do
             Log(string.format("    ModFile: %s\n", ModFile.__name))
             if ModFile.__name == "config.lua" then
                 dofile(ModFile.__absolute_path)
+                -- can someone explain what logic below is supposed to do or under which condition it will ever run
```
UE4SS, line 185: This is to allow changing the asset name from the default `ModActor_C` via a `config.lua` file next to the pak file.
This also requires that the pak file be placed in its own subdirectory in `LogicMods`, instead of directly in `LogicMods`.

Please either remove your comment, or update it with the correct information stated above.

assets/Mods/BPModLoaderMod/Scripts/main.lua
```diff
@@ -110,26 +140,49 @@ local function SetupModOrder()
     end
 end
 
+local function CacheAssetRegistry()
+    ---@cast AssetRegistryHelpers UObject
+    if (AssetRegistryHelpers:IsValid() and AssetRegistry:IsValid()) then return end
+
+    local AssetRegistryHelpers = StaticFindObject("/Script/AssetRegistry.Default__AssetRegistryHelpers") --[[@as UObject]]
```
UE4SS, line 147: Did you even test these changes ?
Making this is a local variable isn't the correct thing to do there.
It has to be global, because we use that global later.
Since this was made local here, the global is always an invalid object, and thus we can't load any BP mods because AssetRegistryHelpers is invalid.
This really should've been caught by local testes before the PR was made.

I also want to add that by moving this function, it appears as entirely new code in the diff which means this change wasn't easy to spot.
It required re-reviewing the entire function instead of just the actual change.

UE4SS (changes requested):
What a mess of a PR.

Parenthesis added to a lot of if-statements where none is needed.

Re-organizing the file in a commit that also has logic changes, causing the diff to be unnecessarily large which in turn requires a full re-review of the moved code to figure out if the code was actually changed (it was) or just moved.

For future PRs, if you cannot avoid reorganizing code, please do so in a separate commit to make it easier to review.

assets/Mods/BPModLoaderMod/Scripts/main.lua
Kein: Can you show an example where this logic actually runs? Assuming hot reload. We are coming into this loop with `Mods {}` empty.

assets/Mods/BPModLoaderMod/Scripts/main.lua
Kein: > Did you even test these changes ?
Yes, it works perfectly, tested by multiple people on OBRE
However, you are right, this should not be `local`. A bit odd it never was an issue.

UE4SS:
> > What a mess of a PR.
> 
> Then simply close it, why are you wasting your time on complaining and/or wasting my time as well.

How is it a waste ?
I provided clear feedback.
If you choose to not take that feedback on, and not even attempt to counter any of it, then that's your problem and you're the one wasting everyones time.

assets/Mods/BPModLoaderMod/Scripts/main.lua
UE4SS: The `config.lua` file is executed in the context of where `dofile` is called.
The idea is for `config.lua` to set its own entry in the `Mods` table, like this:
```lua
-- config.lua
local ModName = "MyTestMod"
Mods[ModName] = {}
Mods[ModName].AssetPath = "/Game/Mods/MyTestMod/CustomModActor"
Mods[ModName].AssetName = "CustomModActor_C"
```

UE4SS (approved):

narknon:
I'd like to cleanup some of the typos/comments before merge, but I can do that if Kein doesn't want to.

Kein:
I can do that but ideally I also would like to either merge `BPML GenericFunctions` to modloader unit or export `BPML GenericFunctions` logic into a shared helpers and use them conditionally in BPModLoader.

My main concern is pointless 2 events that still being registered when no supplementary blueprint mods exist. I'm bad at reading C source but my understanding the match for event name is checked every single time ANY event execution call is, well, called.

P.S. I know that any BP can use these events not just modactors

narknon:
Yeah, BP hooks are rough on performance, though it's just an int comparison. People do use the print even without using our modloader so we can't really just make it suddenly not work, but we have other hooks that seem to hook even when they're not used, which is a mess.

Kein:
May be we can move to a shared instance that can be `require`d by the mods and then the mods can request a hook/BP modloader framework init with simple API/fun call? Under the hood, the API will check if hooks were created and if not, they will be. This way we dont even need to create anything pre-emptively, so if no mods for the game plan to use this functionality - none created.

The only disadvantage is that mods need to opt-in now

assets/Mods/BPModLoaderMod/Scripts/main.lua
```diff
@@ -1,24 +1,43 @@
+---########################
+--- DEFINES
+---########################
+package.path = '.\\Mods\\ModLoaderMod\\?.lua;' .. package.path
+package.path = '.\\Mods\\ModLoaderMod\\BPMods\\?.lua;' .. package.path
```
narknon, line 5: What are these? We don't have any of these paths.

assets/Mods/BPModLoaderMod/Scripts/main.lua
```diff
@@ -1,24 +1,43 @@
...
--- Contains mod names from Mods/BPModLoaderMod/load_order.txt and is used to determine the load order of BP mods.
-local ModOrderList = {}
-
-local DefaultModConfig = {}
-DefaultModConfig.AssetName = "ModActor_C"
-DefaultModConfig.AssetNameAsFName = UEHelpers.FindOrAddFName("ModActor_C")
+--- get table count, never rely on #t
+--- @param T table
+--- @return integer
+local function GetModCount(T)
```
narknon, line 34: (T) is not used here.

assets/Mods/BPModLoaderMod/Scripts/main.lua
```diff
@@ -266,36 +291,57 @@ local function LoadMods(World)
...
+LoadModConfigs()
+LogOrderedMods()
+
+--- Only execute/add any hooks if we have at least one mod entry
+--- It does not matter we can do it manually on a hotkey, PAKs arent mounted
+--- automatially at runtime on LUA reload and we dont need potentially unstable and moot hooks
+if (GetModCount(Mods) > 0) then
+
+    --- future signal to BPML_GenericFunctions we want custom hooks for BP<->lua interaction
+    ModRef:SetSharedVariable("BPModLoaderMod_bIsActive", true)
```
narknon, line 308: If this isn't used yet, we shouldn't include it.

narknon (changes requested):

assets/Mods/BPModLoaderMod/Scripts/main.lua
Kein: Huh? What do you mean, it is used at the very bottom of the script, in MAIN LOGIC session, I pass the `Mods` to get valid count.

assets/Mods/BPModLoaderMod/Scripts/main.lua
Kein: And never will be, just a leftover from my previous experiment. We can't load BPML after ModLoader, there isn't a way (otherwise just started ModActors wont be able to use it)

Kein:
As a proposal, I also changed the reload hotkey to ALT+INS.
This has been something I wanted to propose from the get go but it slipped my mind. Too many games and too many users use INS either in-game (actual bind or a debug bind) or outside (overlay, etc). Reloading properly loaded mods by mistake in-game can cause catastrophic issues.

Xebeth:
I second that. Usually INS is my Reshade key.

UE4SS:
> As a proposal, I also changed the reload hotkey to ALT+INS. This has been something I wanted to propose from the get go but it slipped my mind. Too many games and too many users use INS either in-game (actual bind or a debug bind) or outside (overlay, etc). Reloading properly loaded mods by mistake in-game can cause catastrophic issues.

This should be added to the changelog.

Kein:
Changelog where exactly? The CHANGELOG file?

UE4SS:
> Changelog where exactly? The CHANGELOG file?

`assets/Changelog.md`

PR #900, opened 2025-06-04
Fixes UTC crash timestamp to local time

Kein:
Because the amount of people capable of building of UE4SS can be counted by the fingers of my left hand (and I have all the fingers there) - I cannot test this.

### DIFF
--- a/deps/first/DynamicOutput/src/OutputDevice.cpp
+++ b/deps/first/DynamicOutput/src/OutputDevice.cpp
@@ -25,7 +25,7 @@ namespace RC::Output
 
     auto OutputDevice::get_now_as_string() -> const File::StringType
     {
-        auto now = std::chrono::system_clock::now();
+        auto now = std::chrono::current_zone()->to_local(std::chrono::system_clock::now());
         const File::StringType when_as_string = fmt::format(STR("{:%Y-%m-%d %X}"), now);
         return when_as_string;
     }


narknon:
We shouldn't get the time zone on every single log message, that's very expensive. This would also crash if the timezone database isn't available.

Kobi-Blade:
> We shouldn't get the time zone on every single log message, that's very expensive. This would also crash if the timezone database isn't available.

That is not the only issue, relying on local system time introduces variability that makes it harder to correlate logs or events across different machines and locations.

UTC is the standard for coding, logging, and debugging.

UE4SS:
Additional changes were requested via Discord, but have not been implemented.
These were the requests, and they must be implemented before this gets merged:

1. Two changes in UE4SSProgram.cpp, line 202, and line 478, those are for the GUI console and the non-GUI console.
2. One change in CrashDumper.cpp, line 29, that's for the .dmp file name. 
3. Cache timezone by using the `static` keyword instead of always calling `current_zone()`.

Regarding @Kobi-Blade: I'm unsure how useful timestamps in logs across machines are.
Would an ini setting that reverts to using UTC be an acceptable solution ?

Kobi-Blade:
> Additional changes were requested via Discord, but have not been implemented. These were the requests, and they must be implemented before this gets merged:
> 
> 1. Two changes in UE4SSProgram.cpp, line 202, and line 478, those are for the GUI console and the non-GUI console.
> 2. One change in CrashDumper.cpp, line 29, that's for the .dmp file name.
> 3. Cache timezone by using the `static` keyword instead of always calling `current_zone()`.
> 
> Regarding @Kobi-Blade: I'm unsure how useful timestamps in logs across machines are. Would an ini setting that reverts to using UTC be an acceptable solution ?

Offering an .ini setting to optionally revert to UTC sounds like a reasonable compromise.

narknon:
I will say using UTC has allowed me to tell when someone in a different timezone was sending the wrong/an older log or dmp

UE4SS:
Superseded by #974

UE4SS:
All three points have been addressed, except the option for enabling UTC again for the following reasons:
1. Uncertainty of whether we should because of what narknon said.
2. Separation of modules, causing problems in passing the ini value to the different modules that need it.

PR #902, opened 2025-06-05
Feature/format error msg

Xebeth:
**Description**

**Type of change**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other... Please describe:

**How has this been tested?**

**Checklist**

**Screenshots**

**Additional context**

### DIFF
--- a/deps/first/File/src/FileType/WinFile.cpp
+++ b/deps/first/File/src/FileType/WinFile.cpp
@@ -6,6 +6,7 @@
 
 #define NOMINMAX
 #include <Windows.h>
+#include <Helpers/String.hpp>
 #include <Helpers/Win32Error.hpp>
 #ifdef TEXT
 #undef TEXT
@@ -32,18 +33,16 @@ namespace RC::File
         {
             if (DeleteFileW(file_path_and_name.wstring().c_str()) == 0)
             {
-                auto err = GetLastError();
-                THROW_INTERNAL_FILE_ERROR(fmt::format("[WinFile::delete_file] Was unable to delete file, error: 0x{:x} {}",
-                                                      err, win32_error<char>(err).c_str()))
+                THROW_INTERNAL_FILE_ERROR(fmt::format("[WinFile::delete_file] Was unable to delete file, error: {}",
+                                                      to_string(Win32Error(GetLastError())).c_str()))
             }
         }
         else
         {
-            if (DeleteFileA(file_path_and_name.string().c_str()) != 0)
+            if (DeleteFileA(file_path_and_name.string().c_str()) == 0)
             {
-                auto err = GetLastError();
-                THROW_INTERNAL_FILE_ERROR(fmt::format("[WinFile::delete_file] Was unable to delete file, error: 0x{:x} {}",
-                                                      err, win32_error<char>(err).c_str()))
+                THROW_INTERNAL_FILE_ERROR(fmt::format("[WinFile::delete_file] Was unable to delete file, error: {}",
+                                                      to_string(Win32Error(GetLastError())).c_str()))
             }
         }
     }
@@ -75,7 +74,7 @@ namespace RC::File
 
     auto WinFile::get_raw_handle() noexcept -> void*
     {
-        return static_cast<void*>(m_file);
+        return m_file;
     }
 
     auto WinFile::get_file_path() const noexcept -> const std::filesystem::path&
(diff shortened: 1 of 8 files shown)


deps/first/File/src/FileType/WinFile.cpp
```diff
@@ -168,10 +173,8 @@ namespace RC::File
             {
                 return false;
             }
-            else
-            {
-                deserialize_identifying_properties();
-            }
```
UE4SS, line 174: Why ?

deps/first/File/src/FileType/WinFile.cpp
```diff
@@ -368,12 +369,10 @@ namespace RC::File
...
         {
-            THROW_INTERNAL_FILE_ERROR(fmt::format("[WinFile::close_file] Was unable to close file, error: {}", GetLastError()))
-        }
-        else
-        {
-            set_is_file_open(false);
+            THROW_INTERNAL_FILE_ERROR(fmt::format("[WinFile::close_file] Was unable to close file, {}", to_string(Win32Error(GetLastError())).c_str()))
         }
+
+        set_is_file_open(false);
```
UE4SS, line 375: Why ?

deps/first/Helpers/src/Win32Error.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include <String/StringType.hpp>
+
+#include "Helpers/Win32Error.hpp"
+
+namespace RC
+{
+    Win32Error::Win32Error(DWORD errorCode)
+    {
+        constexpr auto dwLanguageId = MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT);
```
UE4SS, line 13:
```suggestion
        static constexpr auto dwLanguageId = MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT);
```
If your intention was to make this a compile-time constant, use `static constexpr`.
Non-static constexpr doesn't guarantee it's going to be optimized away at compile-time.
Here's an example of non-static constexpr:
image
And here's static constexpr with the same compile options:
image

deps/first/Helpers/src/Win32Error.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+        assign(errorCode, dwLanguageId);
+    }
+
+    Win32Error::Win32Error(HRESULT hResult)
+    {
+        constexpr auto dwLanguageId = MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT);
```
UE4SS, line 20:
```suggestion
        static constexpr auto dwLanguageId = MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT);
```

UE4SS (changes requested):

deps/first/File/src/FileType/WinFile.cpp
Xebeth: That else is useless

deps/first/File/src/FileType/WinFile.cpp
Xebeth: Ditto, else is useless

UE4SS (approved):

narknon:
I'm still not sure that I love not making this crossplatform from the start, or at least setup in a way that crossplatform functionality could be done later by updating in one place. But I can be overruled if you/UE4SS think there are benefits to doing it this way?

UE4SS:
> I'm still not sure that I love not making this crossplatform from the start, or at least setup in a way that crossplatform functionality could be done later by updating in one place. But I can be overruled if you/UE4SS think there are benefits to doing it this way?

I think it's out of the scope for this PR.
It took non-portable code, and wrapped it in a helper that also uses non-portable code, so this changes nothing as far as portability is concerned compared to before this PR.
We can make it portable later.

narknon:
> I think it's out of the scope for this PR. It took non-portable code, and wrapped it in a helper that also uses non-portable code, so this changes nothing as far as portability is concerned compared to before this PR. We can make it portable later.

I don't 100% agree with you, it still adds additional includes. And merging the PR right after I make a comment and before I have any chance to respond to yours is not something that should happen, even if you're making the assumption that I then agree with you.

nicolas-girod:
IIRC the only include I added is Helpers/String.hpp (windows.h was already included in all the files). I think making the code portable was out of scope too. However, since my changes are contained in a class, it's still a step in the right direction in my opinion. The class could easily be renamed to SysError and the format method made portable. I'll have a look after work.

Xebeth:
IIRC the only include I added is Helpers/String.hpp (windows.h was already included in all the files). I think making the code portable was out of scope too. However, since my changes are contained in a class, it's still a step in the right direction in my opinion. The class could easily be renamed to SysError and the format method made portable. I'll have a look after work.

narknon:
Thanks, yeah, the PR is appreciated and is definitely a step in the right direction.  If you want your work account comment deleted let me know.

Xebeth:
If you're able on discord to delete the log from #issues please but ultimately it's no biggie

PR #905, opened 2025-06-05
Make hook non-persistent for console enabler

Kein:
In 99.999% cases we only need to create console object once since it is also a singleton (on a singleton) and after that, the OnObjectConstrcuted event has no purpose for us.
I replaced the construct object hook with a native ufunction hook we can unhook from after we are done, as well as added a check for an imaginary case of dynamicviewport in case we somehow need to re-create the console over and over (I yet to ever meet a single UE game that dynamically juggles viewports).

We cant really get rid of a hook since there is no real 100% guarantee viewport instance is valid once we are loaded with UE4SS (even though this was the case on every test game I tried)

### DIFF
--- a/assets/Mods/ConsoleEnablerMod/Scripts/main.lua
+++ b/assets/Mods/ConsoleEnablerMod/Scripts/main.lua
@@ -1,26 +1,35 @@
+--########################
+-- DEFINITIONS
+--########################
+-- state
 local UEHelpers = require("UEHelpers")
+local pre, post = -1, -1
+local wasConsoleCreated = false
+-- CONFIGURATION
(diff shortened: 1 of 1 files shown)


UE4SS (approved):

narknon:
Looks good, thank you. 

We aren't really going to enforce this heavily for a while but if you could use Conventional Commit style messages in the future it would be appreciated.

Kein:
What's "Conventional Commit style"? Not familiar with.

narknon:
https://www.conventionalcommits.org/en/v1.0.0/

narknon:
Oh oops, I had selected squash but then I guess when I navigated away and back I didn't realize it swapped back to merge commit.

PR #935, opened 2025-06-26
fix(liveview): add unique IDs to function selectables in function caller

narknon:
**Description**

This PR fixes ImGui ID conflict warnings that occur when multiple functions have the same name in the live view function caller widget. The issue was that ImGui uses the selectable label as an identifier, so duplicate function names would create duplicate IDs. 

The fix appends a unique index to each selectable using ImGui's "##" separator convention, which creates unique IDs without changing the displayed text

**Type of change**

**How has this been tested?**

1. Open the live view and select an object that has multiple functions with the same name (e.g., ExecuteUbergraph)
2. Open the function caller widget for that object
3. Verify that:
- All functions are displayed correctly with their original names
- No ImGui ID conflict warning popups appear
- Function selection works properly for all functions, including those with duplicate names
- The selected function's parameters display correctly

**Checklist**

### DIFF
--- a/UE4SS/src/GUI/UFunctionCallerWidget.cpp
+++ b/UE4SS/src/GUI/UFunctionCallerWidget.cpp
@@ -343,9 +343,12 @@ namespace RC::GUI
                         cache_instance(instance);
                     }
 
-                    for (auto& callable_function : m_callable_functions)
+                    for (size_t i = 0; i < m_callable_functions.size(); ++i)
                     {
-                        if (ImGui::Selectable(callable_function.cached_name.c_str(), callable_function.is_selected))
+                        auto& callable_function = m_callable_functions[i];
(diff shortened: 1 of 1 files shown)


UE4SS (approved):
I would've used `std::ranges::enumerate_view` to get the index but your solution is fine as well.

Example for reference:
```c++
#include <ranges>

for (const auto& [index, callable_function] : std::ranges::enumerate_view(m_callable_functions))
{
    // ...
}
```

narknon:
> I would've used `std::ranges::enumerate_view` to get the index but your solution is fine as well.
> 
> Example for reference:
> ```c++
> #include <ranges>
> 
> for (const auto& [index, callable_function] : std::ranges::enumerate_view(m_callable_functions))
> {
>     // ...
> }
> ```

Hmm guess that would be more modern. Don't think it could be const though

PR #939, opened 2025-07-01
fix: push_objectproperty leaving dangling nil on the stack

### LINKED ISSUE #938: [BUG - experimental-latest] [Lua] KismetSystemLibrary.PrintString
**Branch or Release**
UE4SS – v3.0.1 Beta #0 – Git SHA #1b284a7
https://github.com/UE4SS-RE/RE-UE4SS/tree/06688947d98aaa3a01c71964eb6f144bed602dcb

**Game and Engine Version**
Game:  Cash Cleaner Simulator 1.0.6-E
Engine: 5.4

**Describe the bug**
“StrProperty can only be set to a string or FString” error when calling UKismetSystemLibrary::PrintString from UE4SS

**Description**

Calling the _UKismetSystemLibrary::PrintString_ function from Lua results in a _StrProperty_ type error, even when a valid Lua string is passed as the InString argument. The error message is [push_strproperty] Error: StrProperty can only be set to a string or FString.

This suggests that the UE4SS Lua binding is failing to correctly marshal the Lua string into the expected FString parameter on the call stack. The error occurs at _StoredAtIndex: 1_, which corresponds to the _InString_ parameter, indicating the issue is not with the context object but with the string argument itself.

When attempting to call the PrintString function from Lua via UE4SS in UE5.4, the following StrProperty type error occurs:

text

> [push_strproperty] Error:
>  StrProperty can only be set to a string or FString
>     Property: /Script/Engine.KismetSystemLibrary:PrintString:InString

The C++ signature is declared as:

cpp
```
UFUNCTION(BlueprintCallable, meta=(WorldContext="WorldContextObject"))
static void PrintString(
    const UObject* WorldContextObject,
    const FString& InString,
    bool bPrintToScreen,
    bool bPrintToLog,
    FLinearColor TextColor,
    float Duration,
    const FName Key
);
```
In the auto-generated SDK (Engine.lua and Engine.hpp), the InString parameter corresponds to _StrProperty_, i.e., _FString_. In Lua code, passing a plain string still triggers UE4SS to misplace arguments in the call stack and attempt to write to a _StrProperty_ with a null pointer, causing a crash.

**Mods directory**
Not applicable. The bug is reproducible with a single script provided in the "To Reproduce" section. No external mods are required.

**To reproduce**
Use [UE4SS_v3.0.1-430-g1b284a7](https://github.com/UE4SS-RE/RE-UE4SS/releases/download/experimental-latest/UE4SS_v3.0.1-430-g1b284a7.zip)

Launch the game with UE4SS v3.0.1 attached.

Use the following Lua script in any mod:

```
RegisterKeyBind(Key.F4, function()
    local UEHelpers = require("UEHelpers")
    local KSL = UEHelpers.GetKismetSystemLibrary()
    assert(KSL and KSL:IsValid(), "KSL not valid")
    KSL.PrintString(
      nil,                 -- WorldContextObject (not nil)
      "TEXT",              -- InString: Lua string
      true, true,          -- bPrintToScreen, bPrintToLog
      {0.0,0.66,1.0,1.0},  -- TextColor: Lua table instead of FLinearColor
      2.0,                 -- Duration
      FName("TestKey")     -- Key
    )
  end)
end)
```

In-game, press the F4 key.

See the [push_strproperty] error in the UE4SS log.

**Actual result**
The script crashes with:

> [push_strproperty] Error:
>  StrProperty can only be set to a string or FString
>     Property: /Script/Engine.KismetSystemLibrary:PrintString:InString

**Expected behavior**
The _PrintString_ function should execute successfully, displaying the provided text on the screen and/or in the game log. The Lua string "TEXT" should be correctly marshaled to an _FString_ without causing a type error.

The full UE4SS.log file from game launch until the error occurs should be attached to the issue. A .dmp file is not generated for this type of Lua script error, but the log contains the necessary stack trace.

Image

[UE4SS.log](https://github.com/user-attachments/files/20989712/UE4SS.log)

 - OS: Win10

**Additional context**
The issue appears to be specific to functions expecting an _FString_ (_StrProperty_). Calling KSL._PrintText_, which expects an _FText_ (_TextProperty_), works correctly when passed an _FText_ object.

The error persists regardless of how other arguments are passed. The following variations were tested and all resulted in the same StrProperty error:

Passing nil as the WorldContextObject.

Passing the CDO (KSL) as the WorldContextObject.

Passing FLinearColor as a Lua table (e.g., {0,0,1,1}).

Using FName for the Key parameter.

The problem seems to be isolated to the push_strproperty implementation within the Lua binding's UFunction call handler.

UE4SS:
**Description**

If another pusher is executed after `push_objectproperty` with a nil value, it will now act on that nil value instead of the correct value.

Fixes #938 

**Type of change**

**How has this been tested?**

Tested successfully by the author #938.
They had another irrelevant problem as well, and after they fixed that this PR fixed the main problem.

**Checklist**

### DIFF
--- a/UE4SS/src/LuaType/LuaUObject.cpp
+++ b/UE4SS/src/LuaType/LuaUObject.cpp
@@ -439,7 +439,6 @@ namespace RC::LuaType
             else if (params.lua.is_nil())
             {
                 params.lua.discard_value(params.stored_at_index);
-                params.lua.discard_value();
             }
             else
             {


narknon:
It looks like this issue may exist with interface properties too?

UE4SS:
> It looks like this issue may exist with interface properties too?

Yes, and also class properties.
Both have been fixed now.

narknon (approved):

PR #982, opened 2025-08-26
refactor(lua): create struct conversion helpers from push_structproperty

narknon:
**Description**

Creates reusable convert_lua_table_to_struct and convert_struct_to_lua_table helper functions
- Extract convert_lua_table_to_struct and convert_struct_to_lua_table from push_structproperty to helper functions
- Add special handling for arrays with unsupported inner types
- Replace inline lambdas

**Type of change**

**How has this been tested?**

Tested in a couple of games. More testing requested

**Checklist**

**Screenshots**

**Additional context**

### DIFF
--- a/UE4SS/src/LuaType/LuaUObject.cpp
+++ b/UE4SS/src/LuaType/LuaUObject.cpp
@@ -504,6 +504,165 @@ namespace RC::LuaType
         push_integer<uint64_t>(params);
     }
 
+    /**
+     * Helper function to convert a Lua table to a UScriptStruct in memory
+     * @param lua The Lua state
+     * @param script_struct The struct type to convert to
+     * @param data Pointer to the memory where the struct should be written
+     * @param table_index Stack index where the Lua table is located
+     * @param base Optional UObject base for context
+     */
+    auto convert_lua_table_to_struct(
+            const LuaMadeSimple::Lua& lua,
+            Unreal::UScriptStruct* script_struct,
+            void* data,
+            int table_index,
+            Unreal::UObject* base
+            ) -> void
+    {
+        if (!script_struct || !data)
+        {
+            lua.throw_error("convert_lua_table_to_struct: script_struct or data is null");
+            return;
+        }
+
+        unsigned char* data_bytes = static_cast<unsigned char*>(data);
+
+        for (Unreal::FProperty* field : script_struct->ForEachPropertyInChain())
+        {
+            Unreal::FName field_type_fname = field->GetClass().GetFName();
+            const std::string field_name = to_string(field->GetName());
+
+            // Push the field name (key for the table)
+            lua_pushstring(lua.get_lua_state(), field_name.c_str());
+
+            // Get the value from the table
+            // For lua_rawget, we need the actual stack position of the table
+            // If table_index is positive, it stays the same even after pushing the key
+            // If it's negative, we need to adjust for the key we just pushed
+            
+            // Adjust index for negative values after pushing key
+            int adjusted_index = table_index;
+            if (table_index < 0)
+            {
+                adjusted_index = table_index - 1;
+            }
+
+            auto table_value_type = lua_rawget(lua.get_lua_state(), adjusted_index);
+
+            // If there was nothing in the table for this field, leave default value
+            if (table_value_type == LUA_TNIL || table_value_type == LUA_TNONE)
+            {
+                lua.discard_value(-1);
+                continue;
+            }
+
+            int32_t name_comparison_index = field_type_fname.GetComparisonIndex();
+
+            if (StaticState::m_property_value_pushers.contains(name_comparison_index))
+            {
+                void* field_data = &data_bytes[field->GetOffset_Internal()];
+
+                const PusherParams pusher_params{
+                        .operation = Operation::Set,
+                        .lua = lua,
+                        .base = base ? base : static_cast<Unreal::UObject*>(data),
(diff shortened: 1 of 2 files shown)


UE4SS/src/LuaType/LuaUObject.cpp
```diff
@@ -504,141 +504,191 @@ namespace RC::LuaType
...
-        Unreal::UScriptStruct* script_struct = struct_property->GetStruct();
+        if (!script_struct || !data)
+        {
+            lua.throw_error("convert_lua_table_to_struct: script_struct or data is null");
+            return;
+        }
 
-        auto iterate_struct_and_turn_into_lua_table = [&](const LuaMadeSimple::Lua& lua, void* data_ptr) {
-            auto* data = static_cast<unsigned char*>(data_ptr);
+        unsigned char* data_bytes = static_cast<unsigned char*>(data);
```
UE4SS, line 529: Why is this variable created so far from where it's actually used ?
Actually, why does it exist at all ? Just cast it appropriately in the one place where it's used.

UE4SS/src/LuaType/LuaUObject.cpp
```diff
@@ -504,141 +504,191 @@ namespace RC::LuaType
...
+            ) -> void
+    {
+        if (!script_struct || !data)
+        {
+            lua.set_nil();
+            return;
+        }
 
-            // Duplicating the table and putting the duplicate at the top of the stack
+        unsigned char* data_bytes = static_cast<unsigned char*>(data);
```
UE4SS, line 612: Same thing as `convert_lua_table_to_struct`, this variable shouldn't exist here, or at all.

UE4SS/src/LuaType/LuaUObject.cpp
```diff
@@ -504,141 +504,191 @@ namespace RC::LuaType
...
+            int32_t name_comparison_index = field_type_fname.GetComparisonIndex();
 
-                // Pushing the field name (key for the table)
-                lua_pushstring(params.lua.get_lua_state(), field_name.c_str());
+            // Check if we can handle this field type
+            bool can_handle = StaticState::m_property_value_pushers.contains(name_comparison_index);
 
-                // At the top of the stack now: key to find in table (string)
+            // If it's an array, also check if we can handle the inner type
+            if (can_handle && field->IsA<Unreal::FArrayProperty>())
```
UE4SS, line 629: Doesn't this need to be done for other container types as well ?
Maybe we can work on those in another PR.

UE4SS/src/LuaType/LuaUObject.cpp
```diff
@@ -504,141 +504,191 @@ namespace RC::LuaType
...
+            {
+                lua_table.add_key(field_name.c_str());
 
-                // At the top of the stack now: the value corresponding to table[key] or nil
+                const PusherParams pusher_params{
+                        .operation = Operation::GetNonTrivialLocal,
+                        .lua = lua,
+                        .base = base ? base : Helper::Casting::ptr_cast<Unreal::UObject*>(data_bytes),
+                        .data = &data_bytes[field->GetOffset_Internal()],
+                        .property = field
```
UE4SS, line 646: Is there a missing `stored_at_index` assignment here ?
We had one previously.

UE4SS/src/LuaType/LuaUObject.cpp
```diff
@@ -504,141 +504,191 @@ namespace RC::LuaType
...
-                    continue;
-                }
+                StaticState::m_property_value_pushers[name_comparison_index](pusher_params);
+                lua_table.fuse_pair();
+            }
+            else
+            {
+                // Skip fields without handlers
+                Output::send<LogLevel::Verbose>(
+                        STR("Skipping field '{}' of type '{}' (no handler)\n"),
```
UE4SS, line 656: Missing function name prefix in this log message.
The other function has it, so this one should too.

UE4SS (changes requested):

UE4SS/src/LuaType/LuaUObject.cpp
narknon: stored_at_index should be primarily for setting values, for GetNonTrivialLocal we're just pushing the entire result onto the stack, so it shouldn't be needed.

UE4SS/src/LuaType/LuaUObject.cpp
UE4SS: It's cast to `unsigned char*` first, then the first usage casts it to a `UObject*`, and the second usage doesn't cast it at all.
I think it makes more sense for the variable to not exist, and instead cast `data` to `UObject*`, and `uint8_t*` respectively the one time that either are needed, in both functions.

If you disagree and don't end up doing it, you should at least move the creation of the variable as close to the usage(s) as possible.
Smaller scope is better both for optimization potential (though that might not matter in this case) but also for humans reading the code as they don't have to bother understanding the variable until they get to the area where it's actually used, and they can immediately discard it from their brain when it goes out of scope.

If you create a variable at the top of a function (or otherwise far away from where it's used), then the reader has to assume it will be used throughout the function because they have reduced context to understand what it is, and it has to remain in the brain until the very end of the function.

UE4SS/src/LuaType/LuaUObject.cpp
UE4SS: Ah yes, my bad.

UE4SS (approved):

PR #1016, opened 2025-09-23
Add custom signatures for Deadly Days: Roadtrip

xADDBx:
**Description**
Add some custom overrides for the game Deadly Days: Roadtrip:
- Custom UE4SS-settings.ini for an EngineVersionOverride (as the game is using UE 5.6.1 which doesn't seem to be detected as of now)
- Custom StaticConstructObject signature
- Custom FText::FromString signature (in place of the FText constructor, which I was unable to find in a blank game)

**Type of change**

**How has this been tested?**
- I successfully started the game; StaticConstructObject was found in the Phase 1 scan; FText::FText was found in the Phase 2 scan.


narknon:
Thank you. You may want to see if you can use a shorter AOB and still get a unique result just so the sig is less likely to break in future game updates, but can merge this for now and see if it ever breaks and update it then if so. Up to you.

xADDBx:
> Thank you. You may want to see if you can use a shorter AOB and still get a unique result just so the sig is less likely to break in future game updates, but can merge this for now and see if it ever breaks and update it then if so. Up to you.

I'll try to shorten them tomorrow, so the merge can wait for now I guess

xADDBx:
> Thank you. You may want to see if you can use a shorter AOB and still get a unique result just so the sig is less likely to break in future game updates, but can merge this for now and see if it ever breaks and update it then if so. Up to you.

`StaticConstructObject` was easy enough to shorten. `FText::FromString(FString&&)`, however, needs nearly its whole method as signature. While it would probably be possible to improve that via indirect scanning, this works currently so I think it's okay for now.

narknon:
Got it, thank you.

PR #1026, opened 2025-10-01
feat: Added DoEarlyScan to UE4SS-settings.ini

UE4SS:
**Description**

If set to `0`, the entire first phase will be skipped, restoring behavior what it was before #985

**How has this been tested?**

Test both values in UE4SS-settings.ini, and make sure the `Phase 1` / `Phase 2` messages in UE4SS.log appear appropriately and that no crash occurs.

### DIFF
--- a/UE4SS/include/SettingsManager.hpp
+++ b/UE4SS/include/SettingsManager.hpp
@@ -28,7 +28,7 @@ namespace RC
             int64_t SecondsToScanBeforeGivingUp{30};
             bool UseUObjectArrayCache{true};
             StringType InputSource{STR("Default")};
-            bool DoEarlyScan{true};
+            bool DoEarlyScan{false};
         } General;
 
         struct SectionEngineVersionOverride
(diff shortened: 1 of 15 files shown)


narknon:
Maybe we put this in experimental and default off for now?

UE4SS:
> Maybe we put this in experimental and default off for now?

Off by default sure, but I believe this option isn't really experimental.
I believe it's either going to stay permanently, or be removed entirely, and I don't see the harm in removing an option from the non-experimental section.
Based on that, putting it in the experimental section will just cause harm later if we ever want it to not be experimental.

narknon:
Yeah, that's fine. I do think experimental makes some sense though because over time this feature is going to change pretty significantly as we separate out/untangle some of the init better

UE4SS:
> Yeah, that's fine. I do think experimental makes some sense though because over time this feature is going to change pretty significantly as we separate out/untangle some of the init better

If it changes significantly, we can easily retire this option, and create a new one.
We can't very easily move an experimental option into a non-experimental section.
I'm inclined to suggest that the experimental section should be exclusive to WIP features that will either eventually be rolled into core (with no option to turn off) or decided against and be completely removed.
I can't currently think of a reason why we'd not offer this option or a similar option in the future, so I think it's safer to keep it non-experimental.

narknon:
Fair

Issue #1021, opened 2025-09-27
[BUG - Experimental] Grounded not launching at all after adding UE4SS


clorthax:
**Branch or Release**

UE4SS_v3.0.1-541-gf732ada

**Game and Engine Version**

Grounded 1.4.7.4815, Unreal Engine 4

**Describe the bug**

The game doesn't launch at all after installing ue4ss. It shows up in the background processes tab when I look for it in task manager, and never starts.

**Mods directory**

https://www.mediafire.com/file/3w2ibicnqnc969l/External_Mods.zip/file

Mods are not affecting the issue, I tried to launch it with no external mods, didn't work either.

**To reproduce**

**Expected behavior**

Regular Launch

**UE4SS Log**

[2025-09-27 14:09:26.7327158] Console created
[2025-09-27 14:09:26.7394773] UE4SS - v3.0.1 Beta #0 - Git SHA #f732ada
[2025-09-27 14:09:26.7396164] Timezone: Europe/Berlin
[2025-09-27 14:09:26.7396740] UE4SS Build Configuration: Game__Shipping__Win64 (MSVC)
[2025-09-27 14:09:26.7424275] Setting up mods...
[2025-09-27 14:09:26.7448364] Starting mods (from mods.txt load order)...
[2025-09-27 14:09:26.7452628] Starting mods (from enabled.txt, no defined load order)...
[2025-09-27 14:09:26.7456495] No specific game configuration found, using default configuration file
[2025-09-27 14:09:26.7456915] Config: C:\Games\grounded\Maine\Binaries\Win64\ue4ss\UE4SS-settings.ini

[2025-09-27 14:09:26.7457214] root directory: C:\Games\grounded\Maine\Binaries\Win64\ue4ss
[2025-09-27 14:09:26.7458383] working directory: C:\Games\grounded\Maine\Binaries\Win64\ue4ss
[2025-09-27 14:09:26.7458635] game executable directory: C:\Games\grounded\Maine\Binaries\Win64
[2025-09-27 14:09:26.7459290] game executable: C:\Games\grounded\Maine\Binaries\Win64\Maine-Win64-Shipping.exe (108254920 bytes)

[2025-09-27 14:09:26.7459487] mods directory: C:\Games\grounded\Maine\Binaries\Win64\ue4ss\Mods
[2025-09-27 14:09:26.7459615] log directory: C:\Games\grounded\Maine\Binaries\Win64\ue4ss
[...]

PR #1033, opened 2025-10-09
fix(Log): Crash when chrono::current_zone fails

UE4SS:
**Description**

**Type of change**

**How has this been tested?**

Doesn't crash when using Wine.
Someone please test non-Wine, and on a system where `chrono::current_zone` fails.


UE4SS:
Tested and working when current_zone fails: https://github.com/UE4SS-RE/RE-UE4SS/issues/1021#issuecomment-3383963651

PR #1047, opened 2025-10-19
Remove UEHelpers cache for GetPlayerController and GetWorld

ntsik:
**Description**
The `UEHelpers` caches can become incorrect in some cases. Specifically, I encountered an issue in KH3 where `UEHelpers:GetWorld()` would retrieve the correct `World` instance on the first level load, but after a second level load, the `WorldCache` address now led to something different. Sometimes it would be another valid object, e.g. `Material`, which would then be early returned.

It was considered to fix the cache by adding an `IsA` check. However, this alone wasn't enough to fix the issue. While I did not dig into it too deeply, there were cases where `WorldCache:GetClassPrivate()` appeared to return an invalid pointer and thus would crash inside of `IsA`.

So rather than dealing with the complexity of fixing the caches, it was decided to remove the caches instead as they did not provide much gain.

**Type of change**

**How has this been tested?**

1. Using KH3, install a mod with:

```
RegisterHook("/Game/System/BP_NotifyEventToBlueprint.BP_NotifyEventToBlueprint_C:OnPostLoadMap", function()
	print(UEHelpers:GetWorld():GetFullName())
end)
```

2. Launch the game
3. Wait for the title screen
4. Trigger a level change, e.g. load a save
5. Verify that there is no crash
6. Verify that the printed output has type `World`
7. Repeat from step 2 a few times (since the issue was inconsistent to repro)

**Checklist**

### DIFF
--- a/assets/Changelog.md
+++ b/assets/Changelog.md
@@ -141,6 +141,8 @@ Added custom module searcher with UTF-8 path support for Lua `require()` ([UE4SS
 
 - Added annotation for function `FText` to Types.lua. ([UE4SS #788](https://github.com/UE4SS-RE/RE-UE4SS/pull/788))
 
+- Removed cache for `GetPlayerController` and `GetWorld` [PR #1047](https://github.com/UE4SS-RE/RE-UE4SS/pull/1047)
+
 ### C++ API 
 
 Added `UDataTable` class to C++ API. ([UE4SS #997](https://github.com/UE4SS-RE/RE-UE4SS/pull/997))


ntsik:
Two notes:

- I did not bump the version in `UEHelpers`. My assumption is that this should only be bumped for breaking changes.
- `GetEngine` and `GetGameInstance` caches remain in place. My assumption is that those shouldn't change during normal play and are safe to leave cached.

If my assumptions are wrong, please let me know and I can make the appropriate changes.

UE4SS (approved):

PR #1050, opened 2025-10-22
perf(UEHelpers): Fast path for GetPlayerController

UE4SS:
**Description**

It now tries to get the player controller via GameInstance.LocalPlayers[0], and only falls back on the slow FindAllOf method if that fails.

**Type of change**

- [x] Other... Please describe: Performance improvement

**How has this been tested?**

The following script was run, followed by jumping in Palworld.
No testing has been performed on any other game, or on any server software.
```lua
RegisterHook("/Script/Engine.Character:CanJumpInternal", function(ContextParam)
    local Controller = UEHelpers.GetPlayerController()
    if Controller:IsValid() then
        print(string.format("Controller: %s\n", Controller:GetFullName()))
    end
end)
```


joric:
Much faster now, after the performance-killing change in #1047 (I was getting like 10 fps, instead of 60), but now it also returns debug camera controller (if active). It broke my teleport code ([DebugCamera.lua](https://github.com/joric/SupraTools/blob/main/Scripts/DebugCamera.lua)), because in debug camera mode it returns debug camera controller and there's no pawn there so you'd need a fallback function to get the actual player controller with a pawn.

PR #1058, opened 2025-10-23
Lua UDataTable and container ForEach fixes

narknon:
**Description**

Adds early termination support to ForEach loops for all container types (TArray, TMap, TSet, UDataTable).

Previously, ForEach implementations called `lua.call_function()` with 0 expected return values, meaning callbacks could not signal early termination. This forced users to iterate through entire collections even when they found their target element.

This PR updates all ForEach implementations to:
1. Expect 1 return value from callbacks: `lua.call_function(params, 1)`
2. Check if the return value is boolean `true` to break the loop
3. Properly clean up the Lua stack for non-breaking iterations

The implementation follows the pattern already established by `UEnum::ForEachName`.

**Example usage:**
```lua
-- Stop iterating as soon as target is found
dataTable:ForEachRow(function(rowName, rowData)
    if rowName == "TargetRow" then
        -- Process the target
        targetData = rowData
        return true  -- Break out of the loop
    end
end)
```

**Type of change**

**How has this been tested?**

**Checklist**

**Screenshots**

**Additional context**

### DIFF
--- a/assets/Mods/shared/Types.lua
+++ b/assets/Mods/shared/Types.lua
@@ -1191,7 +1191,7 @@ function UDataTable:GetRowStruct() end
 function UDataTable:GetRowMap() end
 
 ---Finds a row in the DataTable by name
----Returns a UScriptStruct wrapper that provides reference-based access to the row
+---Returns a UScriptStruct that provides reference-based access to the row
 ---Modifications to the returned struct directly affect the DataTable
 ---@param RowName string
 ---@return UScriptStruct?
--- a/docs/lua-api/classes/udatatable.md
+++ b/docs/lua-api/classes/udatatable.md
@@ -23,11 +23,11 @@
 ### GetRowMap()
 - **Return type:** `table`
 - **Returns:** a Lua table containing all rows, with row names as keys and row data as values.
-- Each value is either a struct wrapper or a Lua table depending on whether a custom pusher exists for the row type.
+- Each value is either a UScriptStruct or a Lua table depending on whether a custom pusher exists for the row type.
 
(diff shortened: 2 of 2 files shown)


assets/Mods/shared/Types.lua
```diff
@@ -1188,13 +1191,16 @@ function UDataTable:GetRowStruct() end
 function UDataTable:GetRowMap() end
 
 ---Finds a row in the DataTable by name
+---Returns a UScriptStruct wrapper that provides reference-based access to the row
```
UE4SS:
```suggestion
---Returns a UScriptStruct that provides reference-based access to the row
```
The wrapper is supposed to be transparent to the modder, so we shouldn't mention the internals, it will only cause confusion.

docs/lua-api/classes/udatatable.md
```diff
@@ -0,0 +1,77 @@
...
+- **Returns:** the UScriptStruct that defines the row structure for this data table, or nil if not set.
+
+### GetRowMap()
+- **Return type:** `table`
+- **Returns:** a Lua table containing all rows, with row names as keys and row data as values.
+- Each value is either a struct wrapper or a Lua table depending on whether a custom pusher exists for the row type.
+
+### FindRow(string RowName)
+- **Return type:** `UScriptStruct` | `nil`
+- **Returns:** the row data as a UScriptStruct wrapper if found, or nil if the row doesn't exist.
```
UE4SS:
```suggestion
- **Returns:** the row data as a UScriptStruct if found, or nil if the row doesn't exist.
```

PR #1078, opened 2025-11-08
docs: mention FText::FromString for finding FText in UE5 games

Buckminsterfullerene02:
**Description**

In UE5.6 games (and Nark says UE5.4+ ish) the `FText::FText(FString&)` is not found in blank shipping game with symbols, but UE4SS can resolve `FText::FromString(FString&)` fine instead, which is found and works.

I know this technically isn't a constructor anymore but naming is naming probably should just keep it simple.

**Type of change**

**How has this been tested?**

Created `FText_Constructor.lua` file and it works for a UE5.6 game:
```lua
function Register()
    return "48 8B C4 56 57 48 83 EC 68 48 89 58 18"
end

function OnMatchFound(MatchAddress)
    return MatchAddress
end
```

**Checklist**

**Screenshots**

**Additional context**

### DIFF
--- a/docs/guides/fixing-compatibility-problems.md
+++ b/docs/guides/fixing-compatibility-problems.md
@@ -67,7 +67,7 @@ For more in-depth instructions, see the [advanced guide](./fixing-compatibility-
      Function signature: `public: cdecl FName::FName(wchar_t const * ptr64,enum EFindName) __ptr64`
 - FText_Constructor  (Optional)
   - Must return the exact address of the start of the function 'FText::FText'.  
-    Function signature: `public: cdecl FText::FText(class FString & ptr64)const __ptr64`
+    Function signature: `public: cdecl FText::FText(class FString & ptr64)const __ptr64`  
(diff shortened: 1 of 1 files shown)


docs/guides/fixing-compatibility-problems.md
```diff
@@ -68,6 +68,7 @@ For more in-depth instructions, see the [advanced guide](./fixing-compatibility-
 - FText_Constructor  (Optional)
   - Must return the exact address of the start of the function 'FText::FText'.  
     Function signature: `public: cdecl FText::FText(class FString & ptr64)const __ptr64`
+    If on UE5 and the above signature is not found, try `public: __cdecl FText::FromString(class FString const & ptr64)` instead.
```
UE4SS: This appears on the same line as the previous line, which looks a bit weird.
image

Fix with double space at the end of the previous line:
```suggestion
    Function signature: `public: cdecl FText::FText(class FString & ptr64)const __ptr64`  
    If on UE5 and the above signature is not found, try `public: __cdecl FText::FromString(class FString const & ptr64)` instead.
```

UE4SS (approved):

PR #1080, opened 2025-11-10
fix(LiveView): Don't render values that are too large

UE4SS:
**Description**

Replaces the value in the live view tab for structs that are so large that it becomes unreasonable to render them due to fps loss.

This pertains to very large structs, because ExportTextItem is too slow, causing FPS to drop to less than 1.
Note that the value '500' is arbitrary, please feel free to change this to a more appropriate number if you can find one.

This isn't a solution as much as it is a workaround, and a bad one at that, but it's still better than the entire GUI freezing up because of the slowness of ExportTextItem.

**Type of change**

- [x] Other... Please describe: Performance

**How has this been tested?**

**Checklist**


UE4SS:
I will make the value modifiable in the ini file instead of hard-coded to 500.

UE4SS:
This is ready for review now.

PR #1090, opened 2025-11-19
perf(LiveView): 'Apply filters when not searching' is much faster now

UE4SS:
**Description**

This PRs makes the `Apply filters when not searching` option create a result-set instead of applying each filter during every iteration of GUObjectArray.

This PR also adds a compile-time debugging flag, `RC_LIVE_VIEW_DEBUG_FILTER_RESULTS`, that can be used to dump the reason why an object was filtered out from the live view.
It's off by default, and developers can change the code at around line 1930 in LiveView.cpp if they want to only dump the reasons for specific object(s).
It can be enabled in `UE4SS/include/GUI/LiveView/Filter/SearchFilter.hpp`

**Type of change**

- [x] Other... Please describe: Performance

**How has this been tested?**

Apply some filters, and tick the `Apply filters when not searching` checkbox.
Verify that objects that you expect to find in the results actually appear.

**Checklist**

### DIFF
--- a/assets/Changelog.md
+++ b/assets/Changelog.md
@@ -268,6 +268,8 @@ The following search filters now allow multiple values, with each value separate
 
 Large structs that cause FPS problems no longer render, but you can still click on the struct to render each individual property of the struct. ([UE4SS #1080](https://github.com/UE4SS-RE/RE-UE4SS/pull/1080)) 
 
+Large performance increase when `Apply filters when not searching` is enabled ([UE4SS #1090](https://github.com/UE4SS-RE/RE-UE4SS/pull/1090)) 
+
 ### UHT Dumper 
 
 ### Lua API 


narknon:
So how/when are results updated?

UE4SS:
> So how/when are results updated?

Create/Delete listeners, using the same code as regular text search does, except it bypasses the name filter.

narknon:
Does it really need the RC_LIVE_VIEW_DEBUG_FILTER_RESULTS macros? How useful is it to dump reasons something was filtered? Is it worth the extra code/reduced clarity/reduced debugability of using a macro?

UE4SS:
> Does it really need the RC_LIVE_VIEW_DEBUG_FILTER_RESULTS macros? How useful is it to dump reasons something was filtered? Is it worth the extra code/reduced clarity/reduced debugability of using a macro?

I've had to add similar code multiple times for debugging purposes, so yes I would say it's worth it.

narknon (approved):

PR #1093, opened 2025-11-23
Don't add default mod directory path if overridden in config

atenfyr:
**Description**
I recently noted a bug where, on the latest commit on the "main" branch, the default mod directory path is always added to the list of mod directories, even when the mod directory path overridden in the config using the ModsFolderPath setting. Because it seems to me that the last mod directory is used to read the mods.txt file (?), this means that it is simply not possible to completely override the default Mods directory. It is also not possible to simply remove the default path within the config file, because it is added to the list after the settings file is already parsed. This PR simply modifies the UE4SSProgram::setup_mod_directory_path method to avoid adding the new path altogether when an overridden path is present.

**Type of change**
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other... Please describe:

**How has this been tested?**
You can experimentally verify the bug by moving the local Mods directory to a new location, and overriding the ModsFolderPath to that new location. You should observe that the default Mods directory is still added to the list, and no mods load properly. I tested this on the latest commit on the "main" branch (c4b5add).

With this fix, by default, I observed that behavior is unchanged (the only Mods directory is the default, and mods do load properly from that directory). When the ModsFolderPath setting is changed, I observed that only one Mods directory is logged, which is the new directory that was specified in the config file. Mods seem to load as expected.

**Checklist**

- [] I have added the necessary description of this PR to the changelog, and I have followed the same format as other entries.

I have not updated the changelog because this bug was not present in version v3.0.1. It is a regression.

**Screenshots**
Current behavior is depicted below with ModsFolderPath overridden to a new path. Both paths are read, and UE4SS fails to recognize the existence of "mods.txt" in the overridden path.
image

New behavior is depicted below (overridden)
image
(default/not overridden; note that my Mods directory does not exist in this test, so this operates exactly as expected)
image

**Additional context**
I want to note that this is not the first time that this feature has been essentially completely broken from a bug that went completely unnoticed. More rigorous testing is definitely necessary moving forward to avoid breaking this feature again.

### DIFF
--- a/UE4SS/src/UE4SSProgram.cpp
+++ b/UE4SS/src/UE4SSProgram.cpp
@@ -497,6 +497,9 @@ namespace RC
 
     auto UE4SSProgram::setup_mod_directory_path() -> void
     {
+        // don't add a new path if we already specified one in the config
+        if (m_mods_directories.size() > 0) return;
+
         std::filesystem::path default_mods_path = m_working_directory / "Mods";
 
         // If no paths were added, check legacy location for fallback


UE4SS:
This is not a good solution.
This PR seems to misunderstand how the system is supposed to work.
With this change, no mods from the `Win64/ue4ss/Mods` directory is ever loaded if you have any additional paths in `UE4SS-settings.ini`, which is not the intended behavior.
On top of that, it gives this error if there's an additional path that happens to be invalid: `directory_iterator::directory_iterator: Path not found.: "C:\MyMods"`
It will then cause no mods to be loaded at all, regardless if there are other valid paths.

The correct solution here is to change how we treat a missing `mods.txt` file
Right now, it's obviously treated as a near fatal error, regardless if there's a `mods.txt` in any of the additional paths.
We should change it to look for mods.txt in every `Mods` path.

UE4SS:
I will come up with a better solution.

UE4SS:
Superseded by #1094

atenfyr:
That's fair. You could try changing my code to instead check if the
ModsFolderPath config is set to be empty, then, like was done in 3.0.1.

To me, it doesn't make sense that, if I set the main ModsFolderPath config
setting, the default path is still left completely intact. If I put a plus
sign before the config setting, that does make sense to me (which my PR
doesn't account for, as you say), but without that prefix, I would
intuitively expect the default path to not be used at all when I am
deliberately setting a config variable to manually set the mods path. This
is the behavior from the last version, and I also would not expect this
behavior to suddenly change, because it breaks backwards compatibility. It
is also unintuitive to me that the default path cannot be removed by using
the minus sign notation (what is even the use case for that feature, then)?
It seems to me that it is currently completely impossible to remove the
default path from the list of paths, but there's no technical reason for it
to be that way, and this was possible in the previous version without any
issue. Probably the best way to implement that is to not add the default
path to the list of mod paths unless the ModsFolderPath config setting
(with no plus or minus prefix) is empty, which is the old behavior from the
last version, and which should not break the new feature of being able to
add multiple paths (the user can just leave the ModsFolderPath config
setting blank and still use +ModsFolderPath separately, right?). That's my
assessment, at least, as a layman.

atenfyr

UE4SS:
The `-` not working on the default path is indeed bad, and I'll fix that in my PR.

Regarding the rest of your comment, it appears there was a misunderstanding about what `ModsFolderPath` was supposed to do from my end.
I was under the impression that `ModsFolderPaths` was supposed to replace it, which is why it just gets added to the same array.
I will fix this in my PR as well.

UE4SS:
Both fixes have been applied in #1094
You can check the first comment for a download to test when the github action has finished building.

PR #1094, opened 2025-11-23
fix: Fixes to `ModsFolderPath` and `ModsFolderPaths` ini options

UE4SS:
**Description**

This PR does the following:
1. Fixes mods.txt and enabled.txt from all available mods paths not being parsed.
    Toggleable by 'ParseModsFromAdditionalPaths' in the Overrides section in UE4SS-settings.ini.
    Defaults to being enabled.
2. Fixes the `-` syntax for the `ModsFolderPaths` ini option not working because the logic was applied too early.
3. Fixed the singular `ModsFolderPath` ini option not functioning as intended due to a misunderstanding in how it was supposed to function.

Supersedes #1093 

**Type of change**

**How has this been tested?**

First make sure that `ParseModsFromAdditionalPaths` is set to 1, or no value, or is entirely missing from the ini file.
Then, add one or more additional mod paths in UE4SS-settings.ini, and make sure mods load correctly even if `Win64/ue4ss/Mods/mods.txt` is missing.

**Checklist**

### DIFF
--- a/docs/installation-guide.md
+++ b/docs/installation-guide.md
@@ -170,3 +170,22 @@ set UE4SS_MODS_PATHS=C:\SharedMods;D:\GameMods;E:\TestMods
 ```
 
 Environment variable paths are processed in reverse order (first entry has highest priority). Both INI and environment variable paths support relative and absolute paths, including symlinks.
+
+### Controlling Mods List
+
+By default, UE4SS parses the `mods.txt` file from each mods directory. You can override this to use a single controlling `mods.txt` file:
+
+```ini
+[Overrides]
+; Path to a specific mods.txt file to use as the controlling mod list.
+; If set, ONLY this mods.txt will be parsed instead of mods.txt from all mod directories.
+; Can be relative to working directory or an absolute path.
+ControllingModsTxt = ../MyModsList/mods.txt
+```
+
+This is useful when:
+- Managing mod load order centrally across multiple mod directories
+- Using a shared configuration for multiple game installations
+- Keeping mod lists in version control separate from mod files
+
+When `ControllingModsTxt` is empty (default), UE4SS will parse `mods.txt` from all configured mod directories.


UE4SS:
Converted to draft because more discussion needs to take place before we move forward.

atenfyr:
Everything is looking great from my end. I tested the following cases:

* `ModsFolderPath = `: only default path (good)
* `ModsFolderPath = <new path>`: only new path (good)
* `ModsFolderPath = `: and `+ModsFolderPaths = <new path>`: both default and new path (good), mods.txt is read properly in both default path and new path even if the other does not exist (good)
* `ModsFolderPath = `: and `+ModsFolderPaths = <new path>` and `-ModsFolderPaths = ./Mods`: only new path (excellent)
* `ModsFolderPath = <new path 1>` and `+ModsFolderPaths = <new path 2>`: both new paths (good)
* `ModsFolderPath = <new path 1>` and `+ModsFolderPaths = <new path 2>` and `-ModsFolderPaths = ./Mods`: both new paths (good; no problems when attempting to remove path that doesn't actually exist)
* `ModsFolderPath = `: and `-ModsFolderPaths = ./Mods`: no paths exist, game boots but UE4SS makes no attempt to load any mods (good)

I haven't tested your new `ParseModsFromAdditionalPaths` flag (which I simply left unspecified, so presumably it did indeed default to 1 as expected), but everything else looks good to me.

narknon:
Do we really need the ParseModsFromAdditionalPaths? Shouldn't they all be applied in order, appending unique and setting existing to the enable state of the last parsed?

UE4SS:
> Do we really need the ParseModsFromAdditionalPaths? Shouldn't they all be applied in order, appending unique and setting existing to the enable state of the last parsed?

I added it because I don't like the behavior of parsing mods.txt from all paths.
I prefer only one mods.txt file controlling state and order.
It's defaulted to be on specifically because I know it's likely just me and perhaps other devs that want this kind of control by default.

narknon:
I feel like at that point a setting to specify controlling mods.txt makes more sense than just "bool for whichever mods.txt is first".

UE4SS:
> I feel like at that point a setting to specify controlling mods.txt makes more sense than just "bool for whichever mods.txt is first".

That's a good idea, but my method works fine, and is compliant with the fact that the first entry is always supposed to be the main mods directory, which is either the game directory or the override directory.

I don't feel like implementing your idea myself, but you can of course do it yourself if you feel strongly about it, but I'd ask you to not hold up this PR if it's going to take you a while to do it.
If you do it in a different PR, since this option is extremely niche, the likelihood of someone using it before you obsolete it is negligible, you can probably just completely delete it instead of deprecating it.

narknon:
> > I feel like at that point a setting to specify controlling mods.txt makes more sense than just "bool for whichever mods.txt is first".
> 
> That's a good idea, but my method works fine, and is compliant with the fact that the first entry is always supposed to be the main mods directory, which is either the game directory or the override directory.
> 
> I don't feel like implementing your idea myself, but you can of course do it yourself if you feel strongly about it, but I'd ask you to not hold up this PR if it's going to take you a while to do it. If you do it in a different PR, since this option is extremely niche, the likelihood of someone using it before you obsolete it is negligible, you can probably just completely delete it instead of deprecating it.

Done.

narknon:
Now it's not niche and can be used by mod managers/set per-game.

UE4SS:
Confirmed working.
```ini
[Overrides]
ControllingModsTxt = Mods/mods.txt
```

PR #1115, opened 2025-12-18
fix - EnumProperty pushing for lua type

ghost:
Commit # 1114 seems to only solve the dump issue of usmap,Currently, I greatly need Lua's support for pushing enumproperty


UE4SS:
This seems fine except for the preexisting lack of bounds checking.
If a mod assigns a value that's too large for the enum, the value will be silently truncated.
The alternative is to reject such assignments, and output an error, but I don't think this is a critical problem ?
Will merge soon regardless.

narknon:
FUint8Property does not exist.

narknon:
Uses types that don't exist, uses variable declarations within cases that are skipped and not compilable. Pretty clearly AI output that was just PR'd without checking anything.

PR #1124, opened 2025-12-21
fix: Fixed UnregisterHook (again)

### LINKED ISSUE #1123: [BUG - Experimental] Astroneer: The Lua UnregisterHook function is not working as expected
**Branch or Release**
- UE4SS_v3.0.1-534-g49eda85 (works as expected)
- UE4SS_v3.0.1-535-g4809391 (crashes the game)
- UE4SS_v3.0.1-788-g9a20a8f - experimental-latest

**Game and Engine Version**
Astroneer - UE 4.27.2.0

**Describe the bug**
The Lua function `UnregisterHook` can cause the game to crash (with UE4SS_v3.0.1-535-g4809391).
With the latest experimental version (UE4SS_v3.0.1-788-g9a20a8f), the game no longer crashes, but it seems that the UnregisterHook function is not executed (and it is not logged in the UE4SS.log file).
The release "UE4SS_v3.0.1-534-g49eda85" works as expected.

Note that I do not encounter this problem with all hooked functions. For example, if I hook `"/Script/Engine.PlayerController:ClientRestart"` and I call `UnregisterHook` on it later, the game will not crash (everything is working correctly).
But if I do the same thing on `"/Script/Astro.AstroCharacter:GetPlayController"` or `"/Script/Astro.DeformTool:HandleTerrainTool"` then the game crashes (or, with the latest experimental release, UnregisterHook is not executed).

**To reproduce**
- Edit `UE4SS-settings.ini`, add the engine version:
```ini
[EngineVersionOverride]
MajorVersion = 4
MinorVersion = 27
```

- Create a Lua mod with the code below:
[Test1Mod.zip](https://github.com/user-attachments/files/24281204/Test1Mod.zip)
```lua
-- local hook = "/Script/Engine.PlayerController:ClientRestart" -- this hook works correctly
local hook = "/Script/Astro.AstroCharacter:GetPlayController" -- executed when you respawn and on game started
-- local hook = "/Script/Astro.DeformTool:HandleTerrainTool" -- executed when you equip the "Terrain Tool"

Pre, Post = RegisterHook(hook, function()
    print("hook", Pre, Post, "\n")
    UnregisterHook(hook, Pre, Post)
end)
```
- Run the game.

**.dmp file, UE4SS Log and UE4SS-settings.ini**
[ue4ss.zip](https://github.com/user-attachments/files/24281257/ue4ss.zip)

**Desktop:**
 - OS: Win10

UE4SS:
**Description**

Made the UEPseudo callback ids sequential, which now means we don't need create and map our own sequential ids in UE4SS.

Fixes #1123

**Type of change**

**How has this been tested?**

Register hooks for functions that execute repeatedly, and put a call to print and then UnregisterHook in the callback.
Trigger the hook repeatedly, and check the log file, and confirm the hook only executed one time.
Do this for both `/Script/` functions and `/Game/` functions.

**Checklist**

- [x] Any dependent changes have been merged and published in downstream modules. (Included in this PR)

**Screenshots**

**Additional context**

### DIFF
--- a/UE4SS/src/Mod/LuaMod.cpp
+++ b/UE4SS/src/Mod/LuaMod.cpp
@@ -232,7 +232,8 @@ namespace RC
         // Fetch the data corresponding to this UFunction
         auto& lua_data = *static_cast<LuaUnrealScriptFunctionData*>(custom_data);
 
-        auto remove_if_scheduled = [&] {
+        // Returns true if a hooks were removed.
+        auto remove_if_scheduled = [&] -> bool {
             if (lua_data.scheduled_for_removal)
             {
                 const auto function_name_no_prefix = get_function_name_without_prefix(lua_data.unreal_function->GetFullName());
@@ -254,12 +255,19 @@ namespace RC
                     return elem.get() == &lua_data;
                 });
 
-                return;
+                return true;
+            }
+            else
+            {
+                return false;
             }
         };
 
         // Removes pre & post-hook callbacks if UnregisterHook was called in the pre-callback.
-        remove_if_scheduled();
+        if (remove_if_scheduled())
+        {
+            return;
+        }
 
         auto process_return_value = [&]() {
             // If 'nil' exists on the Lua stack, that means that the UFunction expected a return value but the Lua script didn't return anything


narknon:
I believe you reintroduced the safety issue. In the post-hook, the rest of the hook logic will still run after your check for scheduled_for_removal because there's no early return in that conditional.

UE4SS:
> I believe you reintroduced the safety issue. In the post-hook, the rest of the hook logic will still run after your check for scheduled_for_removal because there's no early return in that conditional.

The first check for removal is just for the pre-hook, which should never be called again at this point, so it should be safe, shouldn't it ?

narknon:
I guess it depends on whether scheduled for removal being set already means we should run any code at all for the post hook or not. If they're trying to unregister within the hook, it wouldn't be set until after the callback. If it's set prior to reaching the post hook whether in the pre or in code running in another thread, I'm not sure what the better expectation would be, but it seems like not running the logic would make more sense

UE4SS:
> I guess it depends on whether scheduled for removal being set already means we should run any code at all for the post hook or not. If they're trying to unregister within the hook, it wouldn't be set until after the callback. If it's set prior to reaching the post hook whether in the pre or in code running in another thread, I'm not sure what the better expectation would be, but it seems like not running the logic would make more sense

In my mind, I thought users would expect the entire hook logic would fire before it's unregistered, but that may not be what others expect.
If you disagree, that's good enough for me, I can move the unregistration of the post hook up to the first branch and return early if so, let me know.

narknon:
I think maybe it should be in both places? So if it's unregistered in Pre or Post it'll just get cleaned up before any other calls into it, either before the post callback to early return, or after it

narknon:
Could ask @Corporalwill123 and @Lyrth and @Okaetsu

UE4SS:
> I think maybe it should be in both places? So if it's unregistered in Pre or Post it'll just get cleaned up before any other calls into it, either before the post callback to early return, or after it

Yes, my bad, this is true, we need to unregister both hooks both at the beginning and the end of the post-hook to cover unregistration in the pre and post callback, if we go your way instead of my way.
And, yes, let's ask the others.

PR #1139, opened 2026-01-02
refactor(FText): Remove FText_Constructor.lua as an override

UE4SS:
**Description**

It's no longer needed because we no longer call the FText constructor directly.

Fixes all problems with finding the FText constructor, and also any vtable problems in ITextData when calling `ITextData::GetDisplayString`

**Type of change**

- [x] Other... Please describe: Refactor

**How has this been tested?**

Confirm that the following script outputs the expected string.
```lua
local AsTextManualCall = UEHelpers.GetKismetTextLibrary():Conv_StringToText("SOME TEXT THAT SHOULDN'T ALREADY EXIST")
print(string.format("AsTextManualCall: %s %s\n", AsTextManualCall:type(), AsTextManualCall:ToString()))
local AsTextViaConstructor = FText("TEXT VIA CONSTRUCTOR")
print(string.format("AsTextViaConstructor: '%s' '%s'\n", AsTextViaConstructor:type(), AsTextViaConstructor:ToString()))
```

**Checklist**

**Screenshots**

**Additional context**

Requires https://github.com/Re-UE4SS/UEPseudo/pull/165

### DIFF
--- a/docs/guides/fixing-compatibility-problems-advanced.md
+++ b/docs/guides/fixing-compatibility-problems-advanced.md
@@ -26,7 +26,7 @@ These AOBs are used to find functions and variables that are critical for UE4SS 
 
 ## High-Level Overview
 
-1. **Identify Which Signatures Are Missing:** Determine which functions or variables UE4SS cannot find (e.g., GUObjectArray, GMalloc, FName or FText constructors).
+1. **Identify Which Signatures Are Missing:** Determine which functions or variables UE4SS cannot find (e.g., GUObjectArray, GMalloc, FName).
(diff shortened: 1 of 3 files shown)


Buckminsterfullerene02:
This is awesome. Docs will need updating too.

UE4SS:
> This is awesome. Docs will need updating too.

Thanks for reminding me.
The docs should be all updated now.

Buckminsterfullerene02:
This works on my 5.6 test game with your provided test code.

UE4SS:
Note: This PR caused a slight problem due to usage of the FText constructor outside `on_unreal_init`.
This previously worked, and no longer does after this PR was merged.

I don't believe there's any way to fix this, but I don't think it should be fixed either.
It is and always was inherently unsafe to interact with UE before `on_unreal_init` is called.
Unfortunately we do have at least one confirmed mod that was affected by this.

PR #1169, opened 2026-01-26
feat(Lua): copying to StructProperty from ScriptStruct userdata

Corporalwill123:
**Description**
Uses a property function to copy the data between the UScriptStructs

**Type of change**

**How has this been tested?**
I have tested by setting a parameter within a hook with another from the same context.

Crash has tested `SpawnPos = ModActor:get().OutputPosition` on the same mod from [this conversation](https://discord.com/channels/1148976194050605066/1153804021748994208/1464779757752160431)

**Checklist**


Corporalwill123:
No, I think if it contains any of the other data structures (Set,Map,Array), Text and String, it stops being a POD, for the other properties they may or may not be a POD dependent on other flags I think

I have seen a {UObject*, FName} struct from a game not have CPF_IsPlainOldData, but a UserDefinedStruct with those same properties also be a POD

Issue #1228, opened 2026-04-01
[BUG - UE4SS_v3.0.1-946, UE5.7.4 Failed to find FName::FName, FUObjectHashTables::Get, StaticConstructObject_Internal, GNatives]


aslavd:
Failed to find FName::FName, FUObjectHashTables::Get, StaticConstructObject_Internal, GNatives: 

zDEV-UE4SS_v3.0.1-946-g265115c0.zip

UE 5.7.4

Game: Way Of The Hunter2

OK, I followed 'Fixing missing AOBs' page. Created an empty project. And it also failed to find these functions. Using debugger and PDB file  I found FName::FName, FUObjectHashTables::Get, StaticConstructObject_Internal byte signatures and put them in 'UE4SS_Signatures' folder, and they work.

But how do we should search for GNatives signature?

UE4SS:
You may not need GNatives, it's optional.
If you're planning on hooking or debugging a lot of functions BP functions, it may end up being useful, but otherwise you can leave the error unfixed.
EDIT: GNatives definitely exists in 5.7, so not sure why you wouldn't be able to find it, perhaps a source build instead of using the epic launcher to install the UE editor would help, or maybe just downloading the debug symbols when installing the editor from the launcher would be sufficient.

FUObjectHashTables is fully unused right now as well, but it will probably be used in the future.

aslavd:
Yes, they are exists.
these are signatures for  UE5.7.4 

**FName_Constructor.lua:**
```lua
function Register()
    return "40 53 48 83 EC 30 48 8B D9 48 89 54 24 20 33 C9 4C 8B CA 44 8B C1 48 85 D2 74 27 0F B7 02 66 85"
end

function OnMatchFound(MatchAddress)
    return MatchAddress
end
```

**GNatives.lua:**
```lua
function Register()
    return "4C 89 63 38 4C 89 63 40 8B 46 38 48 0F BA E0 08 73 05 4D 8B C4 EB 07 4C 63 46 44 4C 03 C7"
	-- it is inside void UObject::SkipFunction(UObject *__hidden this, struct FFrame *, void *const, struct UFunction *)
end

function OnMatchFound(MatchAddress)
    local LeaInstr = MatchAddress - 0x15
[...]

PR #1232, opened 2026-04-14
feat(config): BetterMart 0.5.x+ support (UE 5.7.4)

tajemniktv:
**Description**
Added support for BetterMart 0.5+ versions

**Type of change**

- [X] New game config

**How has this been tested?**
dx64dbg in BetterMart to make sure the pattern matches, tested with the sig files and settings in game. Still missing GNatives and GUObjectHashTables it looks like, but that sounds more confusing than I'm willing to work on this. 

**Checklist**

(If there is a list of game configs, it might require updating)
(Hope so)

**Screenshots**
N/A

**Additional context**
Might still need GNatives. The game is in Early Acess and might change frequently, the update that broke the "default" UE4SS sigs was update to UE5.7.4


tajemniktv:
oh holy, the directories are all messed up. I love working on web version of github... 
lemme clean it up real quick

tajemniktv:
eh, hope it's all good now... will think about GNatives but that's for other PR

tajemniktv:
I just found out it doesn't even require specific game config, it's just UE 5.7.4...
Well, that was fun to work with either way. I'm pretty sure this PR can be closed if that's the case...

UE4SS:
I don't have this game so I'll just trust that you're correct about this config not being needed, however, I still want to mention that some of your changes to UE4SS-settings.ini are not recommended for end-users.
I want to mention this just for future PRs, or perhaps even for your own personal use.

In particular, `ConsoleEnabled`, `GuiConsoleVisible`, and `bUseUObjectArrayCache` are enabled.
These are options are for developers, and they can cause instability and performance problems.
You may or may not want these enabled for yourself, but they can't be enabled in a custom config submitted via PR because these configs are meant to be used by end-users, not developers.

tajemniktv:
Okay now I'm not sure whether it's needed or not... It seems to be the same issue as here: https://github.com/UE4SS-RE/RE-UE4SS/issues/1228
And it looks like the signatures the first commenter posted for 5.7.4 are working fine also for BetterMart. So I'm not sure whether this needs a game specific config or will it be somehow included in UE4SS version? (If it's the first, then yes, that PR is still viable)

Yeah, the dev options are my bad as I copied the settings from my folder, which had UE4SS dev version so yeah...

UE4SS:
If it requires any customizations at all, then it should be a separate custom config.
It's not reasonable to make people guess which one of the existing configs may work with their particular game, so we prefer to have a folder for every game even if the config is identical to another game.

tajemniktv:
Yeah, I guess that makes sense... Then it's probably okay to reopen this PR, although I'm worried that if the game updates game engine again, this config might be outdated (while UE4SS default one might work, if it updates fast enough - See the issue I linked before: It seems that all UE5.7.4 might be affected?)
I actually didn't test it with "default"  settings file (Without engine version fields filled) and signatures for 5.7.4 only. 

So in my perspective: if the UE4SS will provide updated signatures for latest UE versions, then until the game comes out of early access it might be worth keeping using the UE4SS default provided ones. At least until the game comes out of early access. Unless you'd prefer having specific signatures and setting files for every game - although with a risk of them becoming outdated... 

Because as of right probably the only changes needed are using 5.7.4 signatures. 

If more info would be needed, lmk

PR #1244, opened 2026-05-08
build: add support for llvm-ml in cross-compilation toolchains

500Byte:
### What does this PR do?
Adds support for `llvm-ml` in the build system to improve cross-compilation toolchains.

### The Problem
When cross-compiling (especially for Linux or specific toolchains without native MSVC `ml64`), the build system lacked explicit support for the LLVM macro assembler (`llvm-ml`), causing compilation failures or requiring manual build script hacks.

### The Solution
Updated the CMake/build configurations to explicitly support `llvm-ml` as an assembler drop-in, enabling smooth cross-compilation with LLVM-based toolchains.


UE4SS:
This may be a good change, and it's simple so I can forgive the use of AI, though I'm not happy about it.
I haven't tested it yet, and the PR will remain open for the moment.

PR #1245, opened 2026-05-08
fix(LuaMod): Resolve stack indexing and sanitize params in StaticConstructObject

500Byte:
### What does this PR do?
Implements a C++ root-cause fix for memory corruption and access violations related to `StaticConstructObject` in the Lua API.

### The Problem
When passing parameters from Lua to `StaticConstructObject`, the original code used implicit stack advancement to read each argument, which could cause incorrect values to be read if any parameter check failed early.

### The Solution
- Replaced implicit stack reading with explicit index-based access using `lua_gettop()` + positional checks (`lua.is_userdata(N)`, `lua_tointeger(L, N)`).
- Added validation for `InstanceGraph` and `ExternalPackage` to reject invalid non-nil values instead of silently casting them.

### DIFF
--- a/UE4SS/src/Mod/LuaMod.cpp
+++ b/UE4SS/src/Mod/LuaMod.cpp
@@ -1953,16 +1953,34 @@ Overloads:
 
             // Change this to userdata if support for 'FObjectInstancingGraph' is ever added
             void* param_instance_graph{};
-            if (lua.is_integer())
+            if (lua.is_nil())
             {
-                param_instance_graph = reinterpret_cast<void*>(static_cast<uintptr_t>(lua.get_integer()));
+                lua.discard_value();
+            }
+            else if (lua.is_userdata())
+            {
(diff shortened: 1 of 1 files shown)


UE4SS:
First of all, this code is AI generated, important to know, and the PR description itself kind of looks AI generated as well but I can't prove that so I'll give you the benefit of the doubt.

Second, can you clarify (without AI) what the problem actually is ?
Which param can be incorrect, and can you show some code that produces this bug in action ?

Regarding InstanceGraph, and ExternalPackage; They were implemented as integers which are then cast to pointers to allow very advanced usage where the caller knows the address of these objects and would like to use these params even though we don't fully support it.
I believe your change doesn't make sense, not only because of what I said above, but because I don't think there's currently any possible way for someone to get a valid lightuserdata to pass for these params.
Please correct me if I'm wrong.

PR #1231, opened 2026-04-07
feat: OutputDevice build support for Linux

UE4SS:
**Description**

This PR adds some support for DynamicOutput.
I've decided an iterative method of adding Linux support may be more effective than doing it all in one chunk, so I might make more PRs similar to this with small-ish changes for Linux support.

Fixes # (issue) (if applicable)

**Type of change**

- [x] Other... Please describe: Build support for Linux

**How has this been tested?**

This has minimal testing.
Specifically, running UE4SS on Windows has been tested and still works, and using DynamicOutput on Linux has been tested and is working except FileDevice and NewFileDevice. Nothing else has been tested.
The File dependency on Linux is not implemented.

It's tested by making a program that utilizes each dependency independently of UE4SS.
CMakeLists.txt, manually change platforms by un/commenting the line near the bottom of the file.
```cmake
cmake_minimum_required(VERSION 3.22)

set(TARGET TestsForUE4SS)
project(${TARGET})

add_executable(${TARGET} "src/main.cpp")
target_include_directories(${TARGET} PUBLIC "include")

# Check for MSVC-compatible compilers (MSVC or clang-cl)
get_filename_component(COMPILER_NAME "${CMAKE_CXX_COMPILER}" NAME)
if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" OR
        COMPILER_NAME MATCHES "clang-cl" OR
(CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "MSVC"))
    target_compile_options(${TARGET} PUBLIC /Zi)
endif()

target_compile_definitions(${TARGET} PRIVATE
        RC_ASM_HELPER_BUILD_STATIC
        RC_FILE_BUILD_STATIC
        RC_DYNAMIC_OUTPUT_BUILD_STATIC
        RC_SINGLE_PASS_SIG_SCANNER_BUILD_STATIC
        RC_SINGLE_PASS_SIG_SCANNER_STATIC
        RC_UNREAL_BUILD_STATIC
        RC_INPUT_BUILD_STATIC
        RC_LUA_MADE_SIMPLE_BUILD_STATIC
        RC_FUNCTION_TIMER_BUILD_STATIC
        RC_PARSER_BASE_BUILD_STATIC
        RC_INI_PARSER_BUILD_STATIC
        RC_JSON_BUILD_STATIC
        RC_JSON_PARSER_BUILD_STATIC
[...]

PR #1246, opened 2026-05-08
fix(linux): Harden directory iteration to prevent Wine/Proton crashes

500Byte:
### What does this PR do?
Introduces stability improvements for users running RE-UE4SS under Linux via Wine/Proton, specifically fixing an infinite recursion crash during directory iteration.

### The Problem
When running under Wine/Proton, the `Z:` drive mapping (which maps the entire Linux root filesystem) can cause `std::filesystem::directory_iterator` to fall into infinite recursion loops (by aggressively following symlinks or diving infinitely deep into restricted system directories), leading to a hard crash on startup.

### The Solution
- Hardened the `iterate_directory` lambda in `LuaMod.cpp` by explicitly skipping symbolic links (`item.is_symlink()`).
- Added a recursion depth limit (`depth > 7`) to completely prevent infinite iteration loops.
- Simplified and improved the `LogicMods` path resolution to avoid manually constructing directory paths that often fail under Linux filesystems.
- Updated `README.md` to accurately reflect the improved Linux/Proton stability status.

> **Note:** This fix targets users running UE4/5 games under Wine/Proton on Linux, where Wine's `Z:` drive mapping causes filesystem iteration issues. It is independent of ongoing native Linux build support (#1231).


README.md
```diff
@@ -62,16 +62,25 @@ RE-UE4SS supports the following environment variables:
...
+- A build system: either [Ninja](https://ninja-build.org/) (recommended) or MSVC (included with Visual Studio)
+
+## Linux Stability Improvements
+
+RE-UE4SS includes specific hardening for Linux/Proton environments to ensure a smooth experience:
+
+- **Robust Directory Iteration:** Filesystem iterators now skip symbolic links and handle permission errors gracefully, preventing infinite recursion crashes (common with Wine's `Z:` drive mapping).
+- **CPU Efficiency:** The async event loop and main update loop use thread-safe atomic flags and proper sleep intervals when paused, eliminating the 100% CPU busy-wait bug.
+- **Thread-Safe Hooks:** Hook management uses a synchronized mutex system and safe data capture to prevent data races and Use-After-Free crashes during mod hot-reloading.
+- **Improved UTF-8 Support:** Robust path conversion ensures that games installed in directories with non-ASCII characters (accents, CJK, etc.) load correctly.
```
UE4SS, line 83: This doesn't belong in the readme.

UE4SS (commented):
I don't like the idea of moving parts of BPModLoaderMod into core UE4SS.
This part of the PR should be removed.

I'm not happy with disallowing symlinks for paths.
It seems more like a workaround than an actual solution.
A better solution might be to check if the canonical path was already iterated.
That would allow the removal of the arbitrary depth limit of 7, and it would also avoid any hard-coded banned paths, while still allowing symlinks.

UE4SS:
I don't have any other way of contacting you, so this will have to do.

I know all three of your PRs have been created with AI, and that is provable by the existence of AI files in your repo, though you seem to have force pushed those files away.

As with any work done by AI, my confidence is low, in both the code but also you as a contributor.
How do we know that you're competent ? Can you fix potential problems in your PRs to a satisfactory degree, or are we going to have to play phone with whatever AI provider you use ?

Overall, the ideas for these PRs are mostly good, however, the implementations are not.
I'm also not happy that you didn't mention that your changes were made using AI.
I'm very appreciative of the fact that these PRs are relatively small, it makes life easier for us.

PR #1252, opened 2026-05-15
fix: ConsoleEnablerMod sometimes not working

UE4SS:
**Description**

This happens when the game overrides the default console class and uses that class to stop the console from appearing.

**Type of change**

**How has this been tested?**

Tested in Subnautica 2.
Without this patch the console won't open.


joric:
I confirm, used it here https://github.com/joric/subnautica/wiki/#ue4ss

narknon:
Don't love the solution because maybe sometimes the custom console could have some useful changes, but probably fine for our generic mod.

narknon (approved):

PR #1268, opened 2026-06-06
Fix for inheritance for some Lua types

UE4SS:
**Description**

There are a few Lua types that are not correctly inherited from their base types.
There are also some classes that incorrectly called `Super::setup_member_functions` which is problematic because that function is already called by the `construct` function, meaning that we end up with two calls to `setup_member_functions`
This has now been fixed.

For future types, prefer to implement inheritance by calling the appropriate `construct` function when possible.

I also decided to document and enforce some hidden behavior with the UScriptStruct inheritance, because it's rather special, and it's important we keep an eye on it so it doesn't break in the future.

**Type of change**

### DIFF
--- a/UE4SS/include/LuaType/LuaUInterface.hpp
+++ b/UE4SS/include/LuaType/LuaUInterface.hpp
@@ -2,7 +2,7 @@
 
 #include <LuaMadeSimple/LuaObject.hpp>
 #include <LuaType/LuaUObject.hpp>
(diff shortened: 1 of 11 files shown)


UE4SS:
We've had at least one person on the discord server say that this fixes several problems, and a crash for them.

wildcherry2 (approved):

PR #1271, opened 2026-06-11
fix: use-after-free in ensure_str_const string intern pool (dangling string_view keys)

### LINKED ISSUE #1270: [BUG - main] Use-after-free in ensure_str_const: intern pool stores caller's string_view as key (crash after Lua GC)
**Branch or Release**

Built from source at `b872ad1`. Still present on `main` as of 2026-06-11 (`272ce2f`): the pool is keyed by `std::string_view` at [String.hpp#L600](https://github.com/UE4SS-RE/RE-UE4SS/blob/272ce2f/deps/first/Helpers/include/Helpers/String.hpp#L600) and the caller's view is stored as the key at [String.hpp#L615](https://github.com/UE4SS-RE/RE-UE4SS/blob/272ce2f/deps/first/Helpers/include/Helpers/String.hpp#L615).

**Game and Engine Version**

The Isle (Evrima) dedicated server, UE 5.6. The bug is game-agnostic — any game running Lua mods that access UObject members is exposed.

**Describe the bug**

Use-after-free in the string intern pool in `Helpers/String.hpp`:

```cpp
auto inline ensure_str_const(std::string_view input) -> const StringType&
{
    static std::unordered_map<std::string_view, StringType> uestringpool;
    ...
    const auto& [emplaced_iter, unused] = uestringpool.emplace(input, std::move(new_str));
```

`emplace(input, ...)` stores the **caller's** `string_view` as the map key — a pointer into memory the caller owns. The hottest caller is the Lua UObject member-access path:

`__index` / `__newindex` → `prepare_to_handle` → `ensure_str_const(lua.get_string())` ([LuaUObject.hpp#L775](https://github.com/UE4SS-RE/RE-UE4SS/blob/272ce2f/UE4SS/include/LuaType/LuaUObject.hpp#L775))

`lua.get_string()` returns `lua_tostring` memory, which is owned by Lua's GC. Once Lua collects that string, the pool key dangles permanently. Any later `ensure_str_const` call whose input hashes into the same bucket runs an equality compare against the freed memory — `EXCEPTION_ACCESS_VIOLATION` if the page has been decommitted, silent garbage reads otherwise.

Two properties make this nasty in practice:

- The strings most likely to be GC'd are exactly the dynamically-built ones (member names assembled at runtime in Lua), while string literals tend to stay alive in Lua's intern table — so it survives light testing and bites under sustained load.
- The crash fires arbitrarily later, on an unrelated key, in whichever thread happens to probe the poisoned bucket — so dumps look like random memory corruption rather than pointing at the insertion site.

**Mods directory**

Not mod-specific — the defect is in stock UE4SS code. Any Lua mod that reads or writes UObject members can plant a dangling key. (Ours are custom server-side mods; nothing from the Mods folder is required to hit this.)

**To reproduce**

1. Run any Lua mod that accesses UObject members using dynamically-constructed strings (e.g. names built with `string.format`/concatenation), so the Lua strings are GC-eligible after the call.
2. Let the Lua GC run (sustained activity is enough).
3. Keep calling into `ensure_str_const` (any UObject member access) — eventually a lookup probes the bucket holding a dangling key and faults.

There is no minimal deterministic repro on a stock allocator because the freed page is usually still mapped. To make it deterministic, run with PageHeap (`gflags /p /enable <exe> /full`) or ASan — the first bucket collision after a GC pass faults immediately.

**Expected behavior**

Interning a string must copy it: the pool has to own its keys. The read path can stay allocation-free with a transparent hash (`is_transparent`) + `std::equal_to<>` for heterogeneous `find(string_view)`.

Observed live on a production dedicated server on 2026-06-11. Crash dump `UECC-Windows-927DBA734C4020C1674876B6ADE5A076` — it's a full dump of a live game server so I'd rather not attach it publicly, but I'm happy to share it privately on request. Symbolized stack (line numbers are from the `b872ad1` build):

```
VCRUNTIME140!memcpy                          <- reading the dangling key's bytes
MSVCP140 std::_Hash find     (xhash:1596)    <- bucket probe / key compare
RC::ensure_str_const         (String.hpp:606)   uestringpool.contains(input)
prepare_to_handle            (LuaUObject.hpp:759)
__index                      (LuaUObject.hpp:486)
[Lua VM frames]
LuaMod                       (LuaMod.cpp:228)
UFunctionStructs             (UFunctionStructs.cpp:100 / 276)
```

 - OS: Windows Server x64

**Additional context**

Working fix, validated in production (rebuilt `UE4SS.dll` deployed to the affected server 2026-06-11, no recurrence since):

- Key type `std::string_view` → owning `std::string`.
- Transparent hash + `std::equal_to<>` so reads remain heterogeneous `find(string_view)` with no allocation on the hot path.
- `emplace(input, ...)` → `try_emplace(std::move(temp_input), ...)` — owns the key and keeps the racing double-insert (two threads both missing under the shared lock) benign.

PR with this fix incoming.

pointsopittha:
**Description**

`ensure_str_const` in `deps/first/Helpers/include/Helpers/String.hpp` interns strings in a static `std::unordered_map<std::string_view, StringType>` and stores the **caller's** `string_view` as the map key. The hottest caller passes Lua-owned memory (`LuaUObject.hpp` `prepare_to_handle` → `ensure_str_const(lua.get_string())`, i.e. `lua_tostring`), so once Lua's GC collects that string the pool key dangles permanently; any later lookup that probes the same bucket compares against freed memory → `EXCEPTION_ACCESS_VIOLATION`. Full analysis with the symbolized production crash stack in #1270.

This PR makes the pool own its keys:

- Key type `std::string_view` → owning `std::string`.
- Transparent hash (`is_transparent`) + `std::equal_to<>` so the read path stays a heterogeneous `find(string_view)` — no allocation per lookup on the hot path.
- `emplace(input, ...)` → `try_emplace(std::move(temp_input), ...)` — the stored key is the owning copy rather than the caller's view, and two threads racing past the shared-lock miss stay benign (the losing insert is a no-op).

Fixes #1270

**Type of change**

**How has this been tested?**

- The crash was observed live on a production The Isle Evrima (UE 5.6) dedicated server (crash dump `UECC-Windows-927DBA734C4020C1674876B6ADE5A076`; stack in #1270 — `memcpy` ← `xhash` bucket probe ← `String.hpp:606` ← `LuaUObject.hpp` `prepare_to_handle`).
- Rebuilt `UE4SS.dll` with this patch and deployed it to that server on 2026-06-11; no recurrence under the same Lua mod workload that previously triggered it.
- The failure is allocator-timing dependent, so there is no deterministic repro on a stock allocator. With PageHeap (`gflags /p /enable <exe> /full`) or ASan, the unpatched code faults on the first bucket collision after a Lua GC pass; the patched code does not.

**Checklist**

**Additional context**

The same hazard exists for any caller passing a transient buffer — Lua strings are just the most common way to hit it, because the signature (`string_view` in, view stored as key) invites it. Owning keys closes the whole class, not just the Lua path.

### DIFF
--- a/assets/Changelog.md
+++ b/assets/Changelog.md
@@ -82,6 +82,8 @@ Added `FUObjectItem`, `FUObjectArray`, and `TUObjectArray` to MemberVariableLayo
 
 Fixed the ingame console (Tilde, or F10 key) not working for some games - ([UE4SS #1252](https://github.com/UE4SS-RE/RE-UE4SS/pull/1252))
 
+Fixed random crashes (use-after-free) caused by the string intern pool storing dangling pointers to caller-owned strings, most commonly Lua GC'd strings from UObject member access ([UE4SS #1271](https://github.com/UE4SS-RE/RE-UE4SS/pull/1271))
+
 ### Live View 
 Added search filter: `IncludeClassNames`. ([UE4SS #472](https://github.com/UE4SS-RE/RE-UE4SS/pull/472)) - Buckminsterfullerene
 
--- a/deps/first/Helpers/include/Helpers/String.hpp
+++ b/deps/first/Helpers/include/Helpers/String.hpp
@@ -597,13 +597,25 @@ namespace RC
     }
     auto inline ensure_str_const(std::string_view input) -> const StringType&
     {
-        static std::unordered_map<std::string_view, StringType> uestringpool;
+        // The pool key must OWN its characters: callers pass string_views into
+        // caller-managed memory (e.g. Lua GC'd strings via lua_tostring), so a
+        // string_view key dangles after the caller's buffer dies and every later
+        // bucket compare reads freed memory (EXCEPTION_ACCESS_VIOLATION).
+        struct TransparentStringHash
+        {
+            using is_transparent = void;
+            auto operator()(std::string_view value) const -> size_t
+            {
+                return std::hash<std::string_view>{}(value);
+            }
+        };
+        static std::unordered_map<std::string, StringType, TransparentStringHash, std::equal_to<>> uestringpool;
         static std::shared_mutex uestringpool_lock;
 
         // Allow multiple readers that are stalled when any thread is writing.
         {
             std::shared_lock<std::shared_mutex> read_guard(uestringpool_lock);
-            if (uestringpool.contains(input)) return uestringpool[input];
+            if (auto existing_entry = uestringpool.find(input); existing_entry != uestringpool.end()) return existing_entry->second;
         }
 
         auto temp_input = std::string{input};
@@ -612,7 +624,7 @@ namespace RC
         // Stall the readers to insert a new string.
         {
             std::lock_guard<std::shared_mutex> write_guard(uestringpool_lock);
-            const auto& [emplaced_iter, unused] = uestringpool.emplace(input, std::move(new_str));
+            const auto& [emplaced_iter, unused] = uestringpool.try_emplace(std::move(temp_input), std::move(new_str));
             return emplaced_iter->second;
         }
     }


wildcherry2 (approved):

UE4SS:
This PR was made with AI, but the code changes are correct, and small, and the description is unusually concise for AI.
I don't know if I should thank you, or be angry because there's no way we can avoid merging these AI changes.

UE4SS (approved):

PR #1274, opened 2026-06-13
fix!: rename misspelled bEnableSeachByMemoryAddress variable

Maxxxxior:
**Description**

Corrected a typo in the `bEnableSearchByMemoryAddress` variable name, which was previously misspelled as `bEnableSeachByMemoryAddress`. This change ensures consistency across the C++ source code and all associated configuration template INI files.

**Type of change**

- [x] Breaking change: Renamed variable to fix typo
- [x] Other: Corrected a typo in a configuration variable name.

**How has this been tested?**

Verified by performing a global search for the misspelled variable throughout the repository to ensure all instances were covered. Confirmed that the new variable name is correctly reflected in `SettingsManager.cpp` and all template INI files.

**Checklist**

**Screenshots**

**Additional context**

Renamed `bEnableSeachByMemoryAddress` to `bEnableSearchByMemoryAddress` to fix a spelling error. All references in the source code and template INI files have been updated.

### DIFF
--- a/UE4SS/src/SettingsManager.cpp
+++ b/UE4SS/src/SettingsManager.cpp
@@ -87,7 +87,7 @@ namespace RC
         REGISTER_INT64_SETTING(General.SecondsToScanBeforeGivingUp, section_general, SecondsToScanBeforeGivingUp)
         REGISTER_BOOL_SETTING(General.UseUObjectArrayCache, section_general, bUseUObjectArrayCache)
         REGISTER_BOOL_SETTING(General.DoEarlyScan, section_general, DoEarlyScan)
-        REGISTER_BOOL_SETTING(General.SearchByAddress, section_general, bEnableSeachByMemoryAddress)
+        REGISTER_BOOL_SETTING(General.SearchByAddress, section_general, bEnableSearchByMemoryAddress)
         StringType default_exec_method_string{};
         REGISTER_STRING_SETTING(default_exec_method_string, section_general, DefaultExecuteInGameThreadMethod)
         if (String::iequal(default_exec_method_string, STR("ProcessEvent")))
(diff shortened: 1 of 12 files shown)


UE4SS:
While the change itself makes sense, breaking compatibility does not, and we cannot accept this PR as is.

If this typo is to be corrected, the way to do it is to leave the old typo'd option functional so that old settings files still load properly and as expected by users, and introduce a new option with the typo fixed.
Our own settings files in the asset directory should of course use the new fixed option like you've already done with this PR.

Fixing this typo would require careful handling in `SettingsManager.cpp` to make sure the old option (with the typo) is preferred over the new one if both are present to make sure old ini files with this option changed from its default still behave correctly.
We don't have an explicit API for checking if a field exists in the ini file, but putting `REGISTER_BOOL_SETTING` in a try/catch block achieves the same thing.

narknon:
Why would we prefer the old if both were present? But this isn't something I'm super worried about maintaining compatability for. Why is it a config setting rather than a search option?

UE4SS:
> Why would we prefer the old if both were present? But this isn't something I'm super worried about maintaining compatability for. Why is it a config setting rather than a search option?

The only reason we care about compatibility is if someone uses a non-default value for this option.
I didn't explain this very well, but both wouldn't actually be present in the file, though this is possible via user-error.
What would happen if we prioritize the new option if a user has changed the value of the old option to be non-default is we would load the default because that's the behavior for any option that doesn't exist in the ini file.
If they use a non-default value, they are probably doing so for a reason.
This is assuming the user doesn't carefully update their ini file, and I think it's reasonable to assume that this scenario will happen, and prioritizing the old option isn't much hassle for us, and I can't think of a reason not to do it.

UE4SS:
@narknon brought up a good point.
This is leftover code from when this feature was first implemented, and this ini option doesn't actually do anything.
The feature is controlled via the `Match memory address` option available if you right click the search bar.

PR #1292, opened 2026-06-20
fix: Returning self when a field doesn't exist

### LINKED ISSUE #1284: [BUG - Release] LUA: Impossible to check if "userdata" variable has a property or method
**Branch or Release**
Latest experimental

```LUA
local x = {}
print(type(x) .. "\n") -- table
print(type(x.aaa) .. "\n") -- nil
if x.aaa then
  print("x has aaa\n")
else
  print("x does not have aaa\n")
end

local y = FindFirstOf("PlayerController")
-- Assuming FindFirstOf successfully returned a valid UObject,
-- Assuming that UObject does not have property or method "aaa",
print(type(y) .. "\n") -- userdata
print(type(y.aaa) .. "\n") -- userdata - unexpected!
if y.aaa then -- Always true!
  print("y has aaa\n")
else
  print("y does not have aaa\n")
end
```

Currently, attempting to reference non-existing property of UObject instance in LUA returns an invalid UObject and not `nil` (see example above).
Is this behavior by design?
If so, it makes it impossible to check if UObject has a specific property or function.
Such check is performed in the UEHelpers.lua and, as it stands now, is pointless:
```LUA
function UEHelpers.GetPlayerController()
-- ...
-- Pointless check: "Controller.IsPlayerController"
            if Controller:IsValid() and (Controller.IsPlayerController and Controller:IsPlayerController() or Controller:IsLocalPlayerController()) then
                return Controller
-- ...
end
```

UE4SS:
**Description**

For example, `FText("Test").aaa` would return a reference of the same FText that tried to access `aaa` instead of nil

**Type of change**

**Checklist**

### DIFF
--- a/UE4SS/src/LuaType/LuaUObject.cpp
+++ b/UE4SS/src/LuaType/LuaUObject.cpp
@@ -2168,7 +2168,7 @@ Overloads:
             else
             {
                 // Construct an empty object to allow for safe chaining with a validity check at the end
-                LuaType::UObject::construct(lua, nullptr);
+                lua.set_nil();
             }
 
             return;
--- a/assets/Changelog.md
+++ b/assets/Changelog.md
@@ -349,6 +349,9 @@ Large performance increase when `Apply filters when not searching` is enabled ([
 ### UHT Dumper 
 
 ### Lua API 
+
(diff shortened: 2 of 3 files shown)


UE4SS:
This will stay as a draft until the discussion in #1284 is concluded.
The PR will either be amended to cover UObject related types, or it will be merged as-is.

wildcherry2 (approved):

PR #1313, opened 2026-06-28
feat: Add support for using mod name as the main dll name for C++ mods

Okaetsu:
**Description**
Allows C++ mods to use the mod name for their main dll, for example if the mod is called `MyAwesomeMod` then you can use `MyAwesomeMod.dll` as the dll name.

This is a fallback, so it'll default to looking for `main.dll` first like before and then `MyAwesomeMod.dll` if `main.dll` wasn't found.

This change makes it slightly easier to work with C++ mods that want to link to another C++ mod where both mods rely on UE4SS, however load order is still something that needs to be accounted for which means the dependency mod must be loaded first before the other mods that depend on it.

**Type of change**

**Checklist**

### DIFF
--- a/UE4SS/include/Mod/CppMod.hpp
+++ b/UE4SS/include/Mod/CppMod.hpp
@@ -23,7 +23,7 @@ namespace RC
         typedef void (*uninstall_type)(CppUserModBase*);
 
       private:
-        StringType m_dll_filename;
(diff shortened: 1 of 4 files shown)


UE4SS (approved):

UE4SS:
Thanks.

PR #1322, opened 2026-07-02
Add monospace font setting for textboxes in debug GUI

Lyrth:
## Description
Add a setting to make the font in some textboxes in the debugging GUI monospaced: `GuiConsoleMonospaceTextEditors = 1`; applied specifically to GUI console log, Watchers logs, Lua debugger text editors (Debug, Script Editor tabs) and REPL tab input and history.

(also arranged the fonts around a bit.)

## Type of change

## How has this been tested?
`GuiConsoleMonospaceTextEditors` disabled: no change in behavior. Check screenshots below for when it's enabled.
Tested only on Linux but should behave the exact same on Windows.

## Checklist

## Screenshots
<details>
<summary>GuiConsoleMonospaceTextEditors enabled screenshots:</summary>

image
image
image
image
image

</details>

## Additional context
Also thought about making a separate setting to make it apply to the entire debug GUI, but I imagine it to be more tedious since there may be some alignment changes that need to be done in a lot of places, so this only covers the tall textboxes.

**Default setting is disabled** since it can *technically* be breaking, for mods that rely on non-monospaced text alignment in the console. But if that doesn't matter much we could have it enabled by default I guess.


UE4SS/src/SettingsManager.cpp
```diff
@@ -137,6 +136,7 @@ namespace RC
         REGISTER_BOOL_SETTING(Debug.DebugConsoleEnabled, section_debug, GuiConsoleEnabled)
         REGISTER_BOOL_SETTING(Debug.DebugConsoleVisible, section_debug, GuiConsoleVisible)
         REGISTER_FLOAT_SETTING(Debug.DebugGUIFontScaling, section_debug, GuiConsoleFontScaling)
+        REGISTER_FLOAT_SETTING(Debug.DebugGUIUseMonospace, section_debug, GuiConsoleMonospaceTextEditors)
```
UE4SS:
```suggestion
        REGISTER_BOOL_SETTING(Debug.DebugGUIUseMonospace, section_debug, GuiConsoleMonospaceTextEditors)
```
Surely this is supposed to be a bool ?

Lyrth: Yep I gotta change that - didnt notice that when I was copying the line above, and it silently compiles and works fine. My bad

UE4SS: It works because "1" can be coerced into 1.0f by `stof`, so it passes our internal "is float" check, and 1.0f gets converted into `true` by the compiler, though I would think it should generate a narrowing conversion warning.

PR #1327, opened 2026-07-05
Change constexpr to inline in special_invalid_ptr

STmihan:
**Description**
Newer MSVC versions (14.44+) are stricter and reject using reinterpret_cast inside the constexpr function special_invalid_ptr. Since this function is only ever used at runtime anyway, removing the constexpr specifier fixes the following build error:
`error C3615: constexpr function 'special_invalid_ptr' cannot result in a constant expression`

**Type of change**

**How has this been tested?**
Built UE4SS from source and a C++ mod against this header with MSVC 14.44+


UE4SS:
Duplicate of #1326

UE4SS:
Thanks for pointing this out still.

STmihan:
Yeah, my bad, was working on a mod too long and just remembered to make a PR. Glad it fixed

Issue #121, opened 2023-05-15
[BUG - All] Unreal Engine version not detected early enough when VTable override is used


narknon:
**Branch or Release**
All

**Game and Engine Version**
Causes issues in 4.25+

**Describe the bug**
When a VTable override is used, UE4SS does not detect the engine version early enough, causing it to assume the version is 0 and for certain version dependent methods to break.

**To Reproduce**
1. Use a vtable override in any 4.25+ game
2. Attempt to use live view or other features

**Expected behavior**
Not crash, properly detect version.

**Desktop (please complete the following information):**
 - OS: Win10

**Additional context**
A workaround is to manually set the engine version in UE4SS-Settings.ini

UE4SS:
I believe this is fixable by separating the `Initialize` function into two functions.
One function does the scanning, and the other one does the rest.
That way we can call the first function, and then set up the vtable layouts, and then call the second function to finish initializing.

narknon:
I remember it being slightly more annoying than that. Something to do with the timing of the version scan. I think the version is necessary to know before some of the 2nd batch scans but on the PS side the engine version is in the same batch as a bunch of other scans so it's chicken and egg. Ideally we'd make a new batch and scan for engine version/override, and then everything else.

When I was messing around with redoing our virtual initialization I remember it being pretty annoying.

MattR69:
I hit what looks like this exact bug on a different title — wanted to add a concrete repro with hand-verified offset math, in case it's useful alongside the fix already discussed above.

**Environment:** Conan Exiles dedicated server (CL `373004`, UE5.6), UE4SS `Git SHA a1e7f571`.

**Symptom:** Reliable `STATUS_ACCESS_VIOLATION` inside `TypeChecker::store_all_object_types`, at `TypeChecker.cpp:279` (`FName::StaticAlignment_Private = static_cast<FNameProperty*>(field)->GetMinAlignment();`). Fixed entirely by adding `[EngineVersionOverride] MajorVersion=5 / MinorVersion=6` to `UE4SS-settings.ini` — but per this issue, that override shouldn't be necessary for correctness, only for cases where auto-detection genuinely fails.

[...]

PR #1392, opened 2026-08-26
Add Conan Exiles Enhanced EngineVersionOverride config (UE4SS-settings.ini)

MattR69:
Conan Exiles' dedicated server (CL `373004`) runs Unreal Engine 5.6, but `UE4SSProgram.cpp` computes `FField`/`FProperty` vtable offsets before its own engine-version auto-detection completes (see #121), so without an explicit `EngineVersionOverride` it silently uses the legacy pre-4.25 offset formula and crashes `TypeChecker::store_all_object_types` on startup, every time.

This adds a `UE4SS-settings.ini` to the existing `CustomGameConfigs/Conan Exiles Enhanced/` folder (which currently only ships `VTableLayout.ini`) setting `EngineVersionOverride` to `5.6`, mirroring the same fix already present for other UE5.6 titles in this repo (`Vein`, `Satisfactory`, `StarRupture`, `Deadly Days Roadtrip`). Confirmed fixing the crash on a live Conan Exiles dedicated server — full startup, map load, mod load, and sustained gameplay simulation, plus a real client connecting and playing. Independently confirmed a second time directly on the Conan Exiles game client binary itself (`ConanSandbox-Win64-Shipping.exe`, same CL `373004`): crashes identically without this override, and starts clean through to `Event loop start` with it applied — so this fix is not dedicated-server-specific, it's needed (and sufficient) for either binary built from this engine version.

This is independent of whatever fix (if any) lands for the underlying ordering bug tracked in #121 — it's a targeted, low-risk workaround for this one game, in the same style this repo already uses for several others.

The file content is the repo's own stock default `UE4SS-settings.ini` template (matching `Deadly Days Roadtrip`'s file byte-for-byte in structure) with only `MajorVersion`/`MinorVersion` under `[EngineVersionOverride]` set — consistent with how the other UE5.6 entries in `CustomGameConfigs` are built, so nothing here is a novel format decision.


UE4SS:
This appears to not be based on a release version of UE4SS-settings.ini which can be found in the non-zDEV zip file in the experimental-latest release.
These custom configs are meant for end-users, so we therefore don't merge configs not based on our official release assets.
Please use UE4SS-settings.ini from the experimental-latest release, and apply your changes on top of that.

MattR69:
Thanks for the pointer — this is now rebuilt from the actual `experimental-latest` release asset (`UE4SS_v3.0.1-1088-ga1e7f571.zip`, the same build this deployment runs), with only `MajorVersion`/`MinorVersion` set under `[EngineVersionOverride]`. Every other value is the untouched release default.

UE4SS (approved):

UE4SS:
Thanks.
I suspect you're using AI to handle all your communications.
If you are, I ask you to stop doing that, we want to communicate with you, the person, not a robot.
If you're not, my apologies.

MattR69:
Fair. I do use AI to write these up (i'm not familiar with the codebase or C++), but the testing and decisions are mine. I'll try to keep it casual moving forward. Thanks for merging it.

PR #1410, opened 2026-09-03
Add Grounded custom game config

### LINKED ISSUE #624: [BUG - Experimental] Blackscreen / Local Character Creation Interruption
**Branch or Release**
The Tag for the Release you are using, or the commit you used if you built from source yourself.
[zDEV-UE4SS_v3.0.1-151-g0e84e2e.zip](https://github.com/UE4SS-RE/RE-UE4SS/releases/download/experimental/zDEV-UE4SS_v3.0.1-151-g0e84e2e.zip)

**Game and Engine Version**
Grounded - 4.27.2.0

**Describe the bug**
When loading into a save the screen is black unless you open the pause menu. I believe somewhere along the line, the character creation is interrupted. It prompts you to pick a teen again which is after you load the save but just before you spawn in.

To be clear, this issue is present even without my mod, i have included my mod in the attachment just because it was asked.

**Mods directory**
[Mods.zip](https://github.com/user-attachments/files/16643981/Mods.zip)

**To Reproduce**
1. Launch game
2. Load into a save game
3. See the blackscreen
4. Press CTRL + F2 to see the error printed from my mod

**Expected behavior**
Usually, the game plays correctly, we have access to statics provided by the game.

[Aidanamitshows the issue here in the discord](https://discord.com/channels/1148976194050605066/1153804021748994208/1273898565114138666)

 - OS: [Win11]

**Additional context**
In the experimental version 137 , the bug is only present when the specific engine version override is set, but in 151 it is present regardless.

@Aidanamite might have additional information.

**Edit:** I found that the error i previously described with getting the local player character returning nil was an issue with my own mod and nothing to do with this issue. I am unsure what actually is the cause now but its still an issue nonetheless. I also have tested version 141 and it doesn't have the blackscreen bug without setting the engine version override.

SSyl:
**Description**
I created a custom config for Grounded (Maine-Win64-Shipping.exe 1.4.7.4815, UE 4.27). The game adds virtuals to AActor, 2 before AddTickPrerequisiteActor and 3 after Tick. Everything in AActor from the tick-prerequisite block through Tick shifts +2 and AGameModeBase shifts +5. In the VTableLayout.ini, FProperty has been padded to its real length.

Without the VTable override, InitGameState resolves to MustSpectate_Implementation (which is the blackscreen/spectator/ghost mode bug), BeginPlay/EndPlay resolve to RemoveTickPrerequisiteActor/Component and never fire.

In UE4SS-settings.ini, I set EngineVersionOverride to 4.27 otherwise a VTable override reads the version as 0 (same as issue #121)

**Fixes**
#624 black screen/spectator/ghost mode issue on world load

**Type of change**
- [x] Other: Custom game config for compatibility

**How has this been tested?**
Tested on UE4SS Experimental 3.0.1-1122-ge4793908 with all 14 hooks enabled.

The Log shows offsets for BeginPlay at 0x348, EndPlay at 0x350, TickActor at 0x418, Tick at 0x470, InitGameState at 0x6C8, and FProperty::GetCPPMacroType at 0x58.

I created a Lua mod that registered all 7 pre/post hooks and printed their class names to the log. Every one fired with the expected classes. InitGameState showed both BP_MainMenuGameMode_C and BP_SurvivalGameMode_C. BeginPlay had 19k fires across appropriate actor classes, as well as EndPlay on teardown.

I also created a C++ mod to test AActorTick's hook to verify it was at the right offset. It fired 5M times with actor contexts.

Just for a sanity check, I also loaded oraxQoL's mod from [Nexus](https://www.nexusmods.com/grounded/mods/107) and it worked in gameplay. The mod itself has 19 UFunctionHooks, so not directly related my changes but helps confirm the changes didn't add regressions.

Offsets were derived by disassembling the shipped vtable.


SSyl:
I also set EnableDumping to 0 because Grounded crashes on exit with 0xc0000005 in ntdll.dll with or without UE4SS, so dumps just pile up at ~45MB each for a crash that UE4SS didn't cause. Confirmed on two of my PCs, one of which has never had UE4SS installed.

narknon:
Thank you!

UE4SS (approved):

SSyl:
Sorry for the late commit. As I was modding Grounded I realized the member variable layout is also different from stock 4.27 in Grounded and not just the vtable. Obsidian adds 8 bytes at the end of UObjectBase so everything in AActor shifts, and they add their own properties on top (ManageTickWithBlendTriggers, bRegisterSceneProxyCulling).

Comparing UE4SS's table against reflection at runtime gave 73/73 mismatches on AActor, 12/12 on AGameModeBase and AGameMode, and 2/5 on UWorld. With MemberVariableLayout.ini in place it comes back clean. UGameViewportClient, UPlayer and ULocalPlayer already matched so they're not in the file.

UWorld only covers its reflected members. TimeSeconds, DeltaTimeSeconds, bInTick and the rest have no UPROPERTY so they aren't in the reflection data.

PR #1412, opened 2026-09-05
fix(lua): guard against null Stack.Node() in script hook dispatcher

Zoinzibar:
**Description**
In `LuaMod::script_hook`, the `execute_hook` lambda unconditionally called `Stack.Node()->GetNamePrivate()` if `callback_container` was non-empty and `precise_name_match` was false.

In certain execution environments (such as synthetic script frames, native thunks, or corrupted execution frames), `Stack.Node()` can return `nullptr`. Dereferencing it caused an immediate access violation crash (`0xC0000005`) in `GetNamePrivate()`.

This PR adds an explicit null check for `Stack.Node()` at the beginning of `execute_hook`, gracefully returning early if no valid `UFunction` node is associated with the script frame, and reuses the guarded local pointer downstream (also eliminating variable shadowing).

**Type of change**

**How has this been tested?**
- Tested in a live game session with multiple active script hooks registered across Blueprint and C++ functions (`BP_PlayerBullet_C:F_StartBullet`, `BP_Enemy_C:F_DamageFeedback`, `BP_Item_C:F_TraceBullet`).
- Confirmed that script hooks continue to fire normally while edge-case frames with null nodes no longer trigger a crash.

**Checklist**

### DIFF
--- a/UE4SS/src/Mod/LuaMod.cpp
+++ b/UE4SS/src/Mod/LuaMod.cpp
@@ -6290,8 +6290,13 @@ Overloads:
             {
                 return;
             }
-            auto data = precise_name_match ? LuaMod::find_function_hook_data(callback_container, Stack.Node())
-                                           : LuaMod::find_function_hook_data(callback_container, Stack.Node()->GetNamePrivate());
+            auto* node = Stack.Node();
+            if (!node)
+            {
+                return;
+            }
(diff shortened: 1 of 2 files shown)


UE4SS (approved):

UE4SS:
Thank you.

PR #1415, opened 2026-09-05
fix(lifecycle): activate FUEDeathListener and pause async tasks during engine shutdown

Zoinzibar:
**Description**
When shutting down Unreal Engine games with active background Lua tasks (such as `LoopAsync` timers or delayed action loops running `FindAllOf` / `FindFirstOf`), clicking "Quit Game" frequently triggered an unhandled access violation crash (`0xC0000005`) in `ntdll.dll`.

### Root Cause
During engine shutdown, Unreal Engine tears down its object tables and critical sections. However, UE4SS background threads (specifically `LuaMod::update_async()`) remained running and continued invoking searcher routines on torn-down engine data structures.

Furthermore, `FUEDeathListener` in `UE4SSProgram.cpp` was previously commented out, preventing `UE4SSProgram::unreal_is_shutting_down` from ever being set when `UObjectArray` shuts down.

### Proposed Changes
- In `UE4SS/include/UE4SSProgram.hpp`:
  - Promote `unreal_is_shutting_down` to `static inline std::atomic<bool>` for thread-safe cross-thread visibility between the shutdown thread and background Lua worker threads.
- In `UE4SS/src/UE4SSProgram.cpp`:
  - Uncomment and register `FUEDeathListener` with `UObjectArray`.
  - On `OnUObjectArrayShutdown()`: set `UE4SSProgram::unreal_is_shutting_down = true` and `UnrealInitializer::StaticStorage::bIsInitialized = false`.
- In `UE4SS/src/Mod/LuaMod.cpp`:
  - `LuaMod::update_async()` now checks `UE4SSProgram::unreal_is_shutting_down` and sleeps rather than executing async actions when the engine is tearing down.
- In `UE4SS/src/main_ue4ss_rewritten.cpp`:
  - In `DLL_PROCESS_DETACH`, set `UE4SSProgram::unreal_is_shutting_down = true`.

**Type of change**

**How has this been tested?**
- Tested with *Far Far West* (UE 5.8) with active background `LoopAsync(500, ...)` enemy searchers:
  - Multiple combat encounters with continuous weapon fire and delayed timers.
  - On clicking "Quit Game" in the main menu: clean exit with 0 access violations and 0 crash dumps generated.

**Checklist**
- [x] I have added the necessary description of this PR to the changelog, and I have followed the same format as other entries (`assets/Changelog.md`).

### DIFF
--- a/UE4SS/include/UE4SSProgram.hpp
+++ b/UE4SS/include/UE4SSProgram.hpp
@@ -94,7 +94,7 @@ namespace RC
 
       public:
         RC_UE4SS_API static SettingsManager settings_manager;
-        static inline bool unreal_is_shutting_down{};
+        static inline std::atomic<bool> unreal_is_shutting_down{};
         static inline std::atomic_bool cpp_mods_done_loading{};
 
       public:
--- a/UE4SS/src/Mod/LuaMod.cpp
+++ b/UE4SS/src/Mod/LuaMod.cpp
@@ -7219,8 +7219,9 @@ Overloads:
     {
         for (m_processing_events = true; m_processing_events && !m_async_thread.get_stop_token().stop_requested();)
         {
-            if (m_pause_events_processing)
+            if (m_pause_events_processing || UE4SSProgram::unreal_is_shutting_down)
             {
+                std::this_thread::sleep_for(std::chrono::milliseconds(20));
                 continue;
             }
 
--- a/UE4SS/src/UE4SSProgram.cpp
+++ b/UE4SS/src/UE4SSProgram.cpp
@@ -70,8 +70,6 @@
 
 namespace RC
 {
-    // Commented out because this system (turn off hotkeys when in-game console is open) it doesn't work properly.
-    /*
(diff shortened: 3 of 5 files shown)


UE4SS:
I don't think this is the correct solution.
I think we should remove `UE4SSProgram::unreal_is_shutting_down` completely, and rely on `Unreal::UnrealInitializer::StaticStorage::bIsInitialized` instead since it's being set in the exact same way as the variable in UE4SSProgram.
Furthermore, if UE is shutting down, we should break out of the loop instead of continuing.

This PR should be split into two:
1. In the UE4SS repo, change all uses of `UE4SSProgram::unreal_is_shutting_down` to `Unreal::UnrealInitializer::StaticStorage::bIsInitialized`, and break any loops instead of continuing when this value is false.
2. In the UEPseudo repo, make ` Unreal::UnrealInitializer::StaticStorage::bIsInitialized` atomic, and make a new delete listener specifically to detect when UE is dying even if `bUseUObjectArrayCache` is false.
I can do the UEPseudo changes if making a PR there is too much of a hassle for you.

PR #1426, opened 2026-09-13
fix: LessEqual421 support for EventViewerMod

Lotruees:
**Description**
Removes plsf hook from LE421 configs as that function was only added in 4.22

Fixes failing to hook into a non existing function in LE421 builds

**Type of change**
- [X] Bug fix (non-breaking change which fixes an issue)

**How has this been tested?**
I tested on a 4.17 game and it now works.


cppmods/EventViewerMod/src/Client.cpp
```diff
@@ -184,19 +184,38 @@ namespace RC::EventViewerMod
                 break;
             }
         }
```
wildcherry2, line 186: You could use ternaries to determine the arrays to use without having to duplicate the rendering logic:
```c++
const bool is_above_421 = RC::Unreal::Version::IsAbove(4,21);
const auto target_name_array = is_above_421 ? EMiddlewareHookTarget_NameArray : EMiddlewareHookTargetLE421_NameArray;
const auto target_name_size = is_above 421 ? EMiddlewareHookTarget_Size :
EMiddlewareHookTargetLE421_Size;
const auto target_value_array = is_above_421 ?  EMiddlewareHookTarget_ValueArray :  EMiddlewareHookTargetLE421_ValueArray;
```
You could also probably make those arrays/size a field of the client class and initialize them in the constructor so that we only compute that once. Since you can't use auto, I believe type of those name arrays would be `const char* const*`

wildcherry2 (changes requested):

PR #1431, opened 2026-09-16
feat: Add new event dispatch system for C++ mods

UE4SS:
**Description**

This is intended to replace the virtual functions in CppUserModBase because we may not be able to add new virtual functions without breaking ABI.

If we create our own dispatcher, we can be fully in charge of the ABI instead of hoping that no toolchains decide to reorder the vtable even if a new virtual function is added at the end.

Even if virtuals added to the end of the vtable was guaranteed to work, it would be annoying to maintain because we'd have to make 100% sure that no new virtual functions get added elsewhere in the class. This new system puts everything behind a pointer to another class which only has function members in it which makes it easy to maintain an ordered list of functions that never change.

**Type of change**


narknon:
We could avoid this breaking abi if we put the callback registry in UE4SS somewhere instead. 
std::unordered_map<CppUserModBase*, ModCallbacks> callbacks;

UE4SS/src/Mod/CppUserModBase.cpp
```diff
@@ -7,12 +7,46 @@
...
+    private:
+        ModEventDispatcher();
+
+    public:
+        // Allocates and returns a ModEventDispatcher.
+        // You must manually free the returned pointer when you're done with it.
+        static auto MakeDispatcher() -> ModEventDispatcher*;
+
+    public:
+        // Member variables must never be removed or reordered.
```
narknon: This is entirely in a .cpp and not visible to mods, so removing/reordering shouldn't matter.

UE4SS:
> We could avoid this breaking abi if we put the callback registry in UE4SS somewhere instead. std::unordered_map<CppUserModBase*, ModCallbacks> callbacks;

Yes, I was worried about the on_update function and doing a hashmap lookup every "update" but maybe it's not that big of a deal since on_update isn't that heavy to begin with.
The only things being done there currently are input event processing, auto reloading Lua mods, and event queue processing which seems to only be for mod management.

UE4SS/src/Mod/CppUserModBase.cpp
UE4SS: Yes this is a leftover comment from an earlier implementation where the members were public and directly accessed.

UE4SS:
I've replaced the pointer with a private singleton.
This fits better than pulling it into UE4SSProgram I think.

UE4SS/src/Mod/CppUserModBase.cpp
```diff
@@ -7,6 +8,38 @@
...
+    {
+    private:
+        ModEventDispatcher();
+
+    public:
+        // Singleton for retrieving (and lazy-constructing) a ModEventDispatcher for the supplied mod.
+        static auto Get(CppUserModBase*) -> ModEventDispatcher&;
+
+    public:
+        // Members may not use STL types.
```
wildcherry2, line 21: Some of these function types use a StringViewType value as a param, which I believe resolves to an STL type. Maybe passing raw const char* or const wchar* would be more stable? Though I guess it would break the API, but something we should do in the future

wildcherry2:
I still think a C API would be simpler/more stable in the long term, but this is still way better than what we have and changing to a C API with a C++ class on top is more trouble than it's worth for now

UE4SS/src/Mod/CppUserModBase.cpp
wildcherry2: Maybe passing it as a `const StringViewType&` could make things a little better at least

UE4SS: Well this is a new system so there's no API to break.
Maybe there's some slight mental overhead when users change to this new system and the types are different though.
I've added a CStringView wrapper which contains the raw pointer as well as the string size.
Can be re-constructed back into a view using the view constructor that takes data + size.
Like this:
```c++
register_on_dll_load([](CppUserModBase* self, CStringView dll_name) {
    Output::send(STR("Loaded dll '{}'\n"), StringViewType{dll_name.data, dll_name.size});
});
```
I think it's a good idea to future proof this system even though the class it lays on top of is a C++ class that also uses C++ types in the form of StringType.
The idea is that we become responsible for this type so we can guarantee ABI compatibility.

wildcherry2: Ah ok, that does preserve the API, misunderstood some stuff. Been doing way too much context switching lately.

Owning the type is definitely the way to go if we must use non-primitive stuff. One extra defensive thing to do would be to static_assert the offsets and sizes of the fields in CStringView, but I wouldn't say it's completely necessary.

PR #1435, opened 2026-09-20
ci: update checkout to v7 in build workflows

katsugtgz:
Updates actions/checkout to the current major (v7) in the PR build workflows.

Changes:
- .github/workflows/cmake-ci.yml: actions/checkout v4 -> v7
- .github/workflows/cmake_build_ue4ss.yml: actions/checkout v4 -> v7

Both files are updated together because cmake-ci.yml calls the reusable build workflow, so bumping only one would leave the pair half-migrated.

Validation:
- both workflow files parse as valid YAML
- checkout v7 has no removed inputs; token and submodules inputs are unchanged

Scope: the two workflow files only. Release and docs deploy workflows are untouched.


katsugtgz:
The Build All Modes run on this PR needs maintainer approval before it starts (first-time contributor workflow). Leaving that to a maintainer, not retrying.

katsugtgz:
Pushed a follow-up to the same branch: cache v4 to v6, upload-artifact v4 to v7, and github-script v7 to v9, so the migration is complete rather than half done. Sorry for the churn on an open PR.

